PR #3, opened 2018-12-14
More efficiently discard inputs to void

dwijnand:


tpolecat:
Uh, how does that work? I haven't seen that one before.

dwijnand:
Deprecated parameters are exempt by the lint warning.

dwijnand:
This change isn't necessary, I just wanted to share.

I'm proposing this function to the standard library.  I'm also proposing `@unused` as an alternative to `@deprecated("unused", "")`.

In experimenting I found out you can remove your need for this function in `StringContextOps` by replacing `void(sc)` with `val _ sc`.  Again, JFYI :)

tpolecat:
It's a good trick to know, thanks!

PR #4, opened 2018-12-15
docker-compose with postgres running on 5432

teigen:
allowing tests and examples to run without setting up
and configuring postgres locally

- creates the 'world' database
- imports the 'world.sql' from doobie


tpolecat:
Thanks!

PR #5, opened 2018-12-21
Upgrade cats-effect and use .allocated for Leak

### LINKED ISSUE #2: Feedback on `Leak`
As requested, a bit of feedback (or maybe a lot, I tend to blabber) on `Leak`, and why the current implementation wouldn't be safe to expose in cats-effect.

Two preliminary points:

- For a possible implementation of `Leak`, this is the approach proposed in cats-effect https://github.com/typelevel/cats-effect/pull/419
- I don't think `Leak` is broken in this codebase, because everything is rewrapped in `Resource.make`, which basically makes it uninterruptible, but I haven't checked too thoroughly.

With that being said, this is the code in question . I've only modified it to use `cf.start { fa }` as opposed to `fa.start`, because it makes things clearer in this case.

```scala
sealed abstract case class Leak[F[_], A](value: A, release: F[Unit])

object Leak {

  private def chain(t: Throwable): Exception =
    new RuntimeException("Resource allocation failed. See cause for details.", t)

  def of[F[_], A](rfa: Resource[F, A])(
    implicit cf: Concurrent[F]
  ): F[Leak[F, A]] =
    for {
      du <- Deferred[F, Unit]
      mv <- MVar[F].empty[Either[Throwable, A]]
      _  <- cf.start {
                  rfa
                      .use {  a => mv.put(Right(a)) *> du.get }
                      .handleErrorWith { e => mv.put(Left(e))  *> du.get }
              }
      e  <- mv.take
      a  <- e match {
              case Right(a) => cf.pure(new Leak(a, du.complete(())) {})
              case Left(e)  => cf.raiseError(chain(e))
            }
    } yield a

}
```

So, let's start highlighting potential problems/improvements.

-----------

 The MVar there is completed at most once, it could be another `Deferred`.

-----------

If `use` fails with an error, which can happen if a step in the resource acquisition fails, `use` guarantees that all the finalisers are run, which means it's unclear why the `handleErrorWith` branch needs to wait on `du.get`. 
  Worse, it sets `mv` to have an error, which means we hit the `Left` branch below, which however does not complete `du`. This means that while resources have been released, we have leaked a Fiber, which is now stuck waiting on `du.get`

------------

  Let's say we successfully `start`, and the resource is acquired correctly, put in `mv`, and waiting for `du.get` to be completed, which will in turn trigger the finalisation. 
     After `start`, there are several points in which the code below it could be interrupted, which means that `du.complete(())` is not called,  `use` doesn't complete, and the finalisers aren't run: this is a resource leak, and the most serious problem. 
   The interruption points are:
           - While waiting on `take`: after the acquisition is started, but before it's completed. `take` (and all operations that involve waiting) introduce async boundaries, which can be interrupted.
           - In cats-effect 1.0.1 (unreleased), in between `flatMaps`, so between `start` and `take`, and between `take` and the `match`.

------------

This is more subtle, but let's take this code using `Resource`:
```scala
    Resource
      .make(IO(println("open")))(_ =>
        IO.sleep(1.second) >> IO(println("close")))
      .use(_ => IO.unit) >> IO(println("next"))
```
The output is this:
```scala
open
close
next
````
 
That is, the code waits until the finalisers are done before moving forward. This is important for two reasons:
1) It's easier to not wait for something designed to wait (via `start`), than to wait for something designed not to wait (synchronisation with extra `Deferred`s, sometimes impossible if not exposed)
2) In cases where you need to release things in order, not waiting introduces non determinism, and possibly closing something still in use by something else.

In such a scenario, even on the happy path the code above can be problematic, because the returned `F[Unit]` finaliser just unblocks the release process by calling `.complete(())`, but doesn't wait for it to finish

------------

Possible solutions: most issues stem from the fact that you need a guarantee that `du.complete(())` will happen no matter what: you need to use `bracket` for that. The final issue with back pressure of finalisers is more subtle and would require extra synchronisation with another `Deferred`, which however means potentially more problems. The approach taken in the cats effect PR instead interprets the Resource ADT, building a chain of `brackets` without relying on concurrency.

tbrown1979:
This PR upgrades the `cats-effect` to `1.1.0`. 

With that, we gain the new method `allocated` for a `Resource`, which is essentially what `Leak` attempted to do. I changed `Leaks` construction to now use this method. I decided to leave `Leak` as a type here because it seems more expressive than just a normal tuple. Let me know if you'd prefer that handled another way 😄 

Fixes #2

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import ReleaseTransformations._
 
 // Library versions all in one place, for convenience and sanity.
 lazy val catsVersion          = "1.4.0"
-lazy val catsEffectVersion    = "1.0.0"
+lazy val catsEffectVersion    = "1.1.0"
 lazy val catsMtlVersion       = "0.3.0"
 lazy val catsTaglessVersion   = "0.1.0"
 lazy val kindProjectorVersion = "0.9.8"
--- a/modules/core/src/main/scala/util/Leak.scala
+++ b/modules/core/src/main/scala/util/Leak.scala
@@ -5,34 +5,14 @@
 package skunk.util
 
 import cats.effect._
-import cats.effect.concurrent._
-import cats.effect.implicits._
 import cats.implicits._
 
 /** A leaked resource and its cleanup program. */
 sealed abstract case class Leak[F[_], A](value: A, release: F[Unit])
 
 object Leak {
-
-  private def chain(t: Throwable): Exception =
-    new RuntimeException("Resource allocation failed. See cause for details.", t)
-
-  def of[F[_], A](rfa: Resource[F, A])(
-    implicit cf: Concurrent[F]
-  ): F[Leak[F, A]] =
-    for {
-      du <- Deferred[F, Unit]
-      mv <- MVar[F].empty[Either[Throwable, A]]
-      _  <- rfa.use {            a => mv.put(Right(a)) *> du.get }
-              .handleErrorWith { e => mv.put(Left(e))  *> du.get }
-              .start
-      e  <- mv.take
-      a  <- e match {
-              case Right(a) => cf.pure(new Leak(a, du.complete(())) {})
-              case Left(e)  => cf.raiseError(chain(e))
-            }
-    } yield a
-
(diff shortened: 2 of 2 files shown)


tpolecat:
Thanks! Sorry for the delay.

PR #6, opened 2019-02-15
Add mapK to Session (and transitively other things)

kubukoz:
Kinda WIP:

- Some traits have `mapK` in the companion to avoid adding extra dependencies for those who don't need it. Not sure how I feel about this inconsistency. They could probably be syntax + "static" defs.
- The `mapK` for Signal is skunk-private for now, but I imagine it'll be moved to fs2.
- Some defs could also be vals.


build.sbt
```diff
@@ -1,12 +1,12 @@
 import ReleaseTransformations._
 
 // Library versions all in one place, for convenience and sanity.
-lazy val catsVersion          = "1.4.0"
-lazy val catsEffectVersion    = "1.1.0"
+lazy val catsVersion          = "1.6.0"
+lazy val catsEffectVersion    = "1.2.0"
 lazy val catsMtlVersion       = "0.3.0"
 lazy val catsTaglessVersion   = "0.1.0"
```
kubukoz: This one seems to be unused.

tpolecat:
I'm working on this again if you feel like rebasing and working this out again, otherwise feel free to close and I'll pick it up again at some point.

tpolecat:
Implemented in #23

PR #44, opened 2019-06-20
Temporal codecs

### LINKED ISSUE #42: Fix temporal codecs (and test them)
Temporal codecs currently do not work too well.

For example:
https://github.com/tpolecat/skunk/blob/346d42bcb3479c7ed9e59ae77e59c79c92d7d1ad/modules/core/src/main/scala/codec/TemporalCodecs.scala#L23
This uses format method on String instead of DateTimeFormatter.

https://github.com/tpolecat/skunk/blob/346d42bcb3479c7ed9e59ae77e59c79c92d7d1ad/modules/core/src/main/scala/codec/TemporalCodecs.scala#L33
I think this isn't quite right, it wouldn't decode e.g. `02:03:06.33` or `02:03:06` and these values are of type `time(6)`. `DateTimeFormatterBuilder` offers quite a nice way to build these, I'm thinking something along these lines:
```scala
private def timeFormatter(precision: Int): DateTimeFormatter = {
    val requiredPart = new DateTimeFormatterBuilder()
      .appendValue(HOUR_OF_DAY, 2)
      .appendLiteral(':')
      .appendValue(MINUTE_OF_HOUR, 2)
      .appendLiteral(':')
      .appendValue(SECOND_OF_MINUTE, 2)

    if(precision > 0) {
      requiredPart.optionalStart()
        .appendFraction(MILLI_OF_SECOND, 0, precision, true)
        .optionalEnd()
        .toFormatter()
    } else {
      requiredPart.toFormatter()  
    }  
  }
```

If I get a green light on this I would love to provide a PR this week.

asoltysik:
TODO:
- [x] interval codec 
- [ ] integration test on real postgres table

Resolves #42


asoltysik:
@tpolecat Ready for review

modules/core/src/main/scala/codec/TemporalCodecs.scala
```diff
@@ -13,83 +13,139 @@ import java.time.LocalDateTime
...
+    } else {
+      throw new IllegalArgumentException(s"timestampz($precision): invalid precision, expected 0-6")
     }
 
-  // todo: intervals, which will need a new data type
+  private def intervalCodec(tpe: Type): Codec[Duration] = Codec.simple(
+    duration => duration.toString,
+    str => Either.catchOnly[DateTimeParseException](Duration.parse(str)).leftMap(_.toString),
+    tpe
+  )
```
tpolecat: Wait, what? Java's `Duration` toString/parse just happen to match Postgres input/output formats?

tpolecat:
Conflict to resolve, sorry. I'm trying to get caught up. This looks good but I'm kind of amazed `Duration` works.

modules/core/src/main/scala/codec/TemporalCodecs.scala
asoltysik: Luckily, it does ;) Duration#toString uses ISO-8601 format, so we have postgres interval input as ISO-8601. We also configure postgres to use ISO-8601 interval output here: https://github.com/tpolecat/skunk/blob/724fb5570e73091406704e6c24a0af4aa395750d/modules/core/src/main/scala/net/message/StartupMessage.scala#L37

asoltysik:
Travis build fails, this seems to be a problem: https://travis-ci.community/t/error-installing-oraclejdk8-expected-feature-release-number-in-range-of-9-to-14-but-got-8/3766/6

modules/core/src/main/scala/codec/TemporalCodecs.scala
asoltysik: Although I don't take the precision into consideration, so somebody writing `sql"'PT5.6666'::interval(6)".query(duration(2))` might expect to get back `PT5.66` but will get `PT5.6666` instead. I will figure out if there was any reason I left it this way and fix if not.

asoltysik: Oh, the above query is actually not possible to execute, because skunk will check that `interval(6)` and `interval(2)` are not the same type and throw an error. So we do not need to trim parts of seconds to right precision when reading from postgres - the precision here will always be right. 

What would still make sense is trimming the strings to the right precision when **writing** to postgres, i.e. when writing with codec `interval(2)` we would trim `PT5.6666` to `PT5.66`. This doesn't change any semantics, just would save a few bytes. Do you think it's worth doing?

tpolecat:
rebased and tidied up a bit in #109

PR #47, opened 2019-07-25
Lint core

lorandszakacs:
Really simple stuff:
- [x] add `override` modifier where appropriate
- [x] add explicit return type to (hopefully) all public class fields/defs
- [x] add final modifiers to case classes


tpolecat:
Thanks! I need to set up scalafix or something.

PR #49, opened 2019-07-25
Add scalafmt

lorandszakacs:
I tried my best to create a `.scalafmt.conf` that best preserves the repo's current style. In roughly 15 places I explicitly turned off formatting with the `// format: off` comment.

My suggestion is that you checkout the branch locally, reset the last commit (the only one that actually applied scalafmt via the `sbt scalafmtAll` task), play around maybe a bit some more with the config until you are happy.

Like any good compromise, everyone is dissatisfied, but at least there is consistent formatting in the project.


lorandszakacs:
Just force pushed because I noticed some areas that shouldn't be formatted.

tpolecat:
Thanks for this, but I don't wish to make this change.

PR #50, opened 2019-07-30
Text changes

ikr0m:
minor changes, typo fixes


tpolecat:
Thanks, some kind of screwup with Oracle JDK ... you can try changing it `travis.yml` to use `openjdk8` instead of `oraclejdk8`. If not I'll track it down.

PR #72, opened 2019-10-19
Fixes insert command

gvolpe:
Insertions like the one below are now working:

```scala
val insertCity: Command[Int ~ String ~ String ~ String ~ Long] =
    sql"""
         INSERT INTO city
         VALUES ($int4, $varchar, $varchar, $varchar, $int8)
       """.command
```

I decided to ignore the parameters in the `Completion` pattern match but I'm not sure if that's correct, I'm not very familiar with the protocol.


tpolecat:
I think the argument is probably an int like the others, reporting the number of modified rows. But I don't know, this part is undocumented.

gvolpe:
No, it's not :smile: ! The arguments indicate the number of arguments we pass to values.

What we get is `INSERT 0`, or `INSERT 0 1`, or `INSERT 0 1 2`, etc. That didn't seem relevant to me and I've no idea of how to parse that using regex hehe

gvolpe:
```
[error] Error: Total 26, Failed 0, Errors 1, Passed 25
[error] Error during tests:
[error] 	tests.CommandTest
[error] (tests / Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 54 s, completed Oct 19, 2019 7:47:41 PM
```

This is strange, tests are passing for me locally. Maybe restart?

gvolpe:
> I think the argument is probably an int like the others, reporting the number of modified rows

Okay, I'm back. I forgot to mention that I also thought that was the case and this was the first thing I tried:

```scala
val Insert: Regex = """INSERT (\d+)""".r
```

Then I discovered that it was failing when I was inserting two values:

```scala
val insertCmd: Command[String ~ String] =
    sql"""
         INSERT INTO brands
         VALUES ($varchar, $varchar)
       """.command
```

This was the error I was getting:

```
scala.MatchError: INSERT 0 1 (of class java.lang.String)
	at skunk.net.message.CommandComplete$.$anonfun$decoder$1(CommandComplete.scala:46)
	at scodec.DecodeResult.map(DecodeResult.scala:17)
	at scodec.Decoder$$anon$1.$anonfun$decode$1(Decoder.scala:48)
	at scodec.Attempt$Successful.map(Attempt.scala:111)
	at scodec.Decoder$$anon$1.decode(Decoder.scala:48)
	at scodec.GenCodec$$anon$2.decode(GenCodec.scala:76)
	at scodec.Decoder.decodeValue(Decoder.scala:41)
	at scodec.Decoder.decodeValue$(Decoder.scala:41)
	at scodec.GenCodec$$anon$2.decodeValue(GenCodec.scala:73)
	at skunk.net.MessageSocket$$anon$1.$anonfun$receiveImpl$2(MessageSocket.scala:54)
```

Then I realized that it returns the number of arguments in the same way so I added other patterns:

```scala
val Insert1: Regex = """INSERT (\d+)""".r
val Insert2: Regex = """INSERT (\d+ \d+)""".r
val Insert2: Regex = """INSERT (\d+ \d+ \d+)""".r
```

This does work but as you can see it gets ugly quickly, and I don't know how to write such a regex. After some Googleing I couldn't figure it out so I wrote a match-all after the INSERT clause since the result doesn't seem relevant anyway.

Hope that clarifies things a bit! :)

gvolpe:
Hey @tpolecat ,

It turns out I was wrong. The pattern is always something like `INSERT 0 1` where the second number indicates the number of records inserted and the first one related to some `OIDS` I have no idea about.

https://www.postgresql.org/message-id/4AD5F063.8050708@iol.ie

I fixed the pattern now and can confirm it works well.

tpolecat:
This isn't going to work if oids are turned on but I'll open an issue and we can follow up. Thanks!

PR #77, opened 2019-10-23
add generic mappings

tpolecat:
This adds functor mappings from twiddle list types `A ~ B ~ ...` to isomorphic case class types. So we have:

- `decoder.gmap[A]`
- `encoder.gcontramap[A]`
- `coder.gimap[A]`

So basically this just simplifies the writing of codecs for case classes. In `CommandTests.scala` this

```scala
val city = (int4 ~ varchar ~ bpchar(3) ~ varchar ~ int4).imap 
  { case i ~ n ~ c ~ d ~ p => City(i, n, c, d, p) }
  { case c => c.id ~ c.name ~ c.code ~ c.district ~ c.pop }
```

becomes this

```scala
val city = (int4 ~ varchar ~ bpchar(3) ~ varchar ~ int4).gimap[City]
```

which seems better.

Because this pulls in shapeless I added it as an optional module. To use it you need to add the `skunk-generic` artifact and `import skunk.generic.implicits._`.

/cc @gvolpe … also thanks @milessabin for help with a spooky inference issue.


gvolpe:
:rocket:

PR #85, opened 2019-11-05
added create/drop table command

lightsaway:
Hi, i've tried to execute create table command, which didn't succeed. 
Added this command + drop table command and tests for it.


tpolecat:
Cool, thanks!

tpolecat:
This patch will be available in `0.0.4+27-057c59c2-SNAPSHOT` which is building right now.

PR #86, opened 2019-11-05
fixing encoding of temporal simple codec

lightsaway:
Current encoder for temporal is not working properly
```scala
object Main{
  def main(args: Array[String]): Unit = {
    val date = LocalDateTime.now()
    println(temporal.timestamp(1).encode(date))
  }
}
```
produces folowing
```
List(Some(yyyy-MM-dd HH:mm:ss.S))
```
I saw that there is some fix temporal branch, didn't look too deep into it. I can drop this one, or if this change is ok, i can add some tests for whole encodings


tpolecat:
Haha nice. Thanks.

PR #130, opened 2020-01-21
Fix update completion

DylanArnold:
Fixes https://github.com/tpolecat/skunk/issues/128


tpolecat:
Thanks! Sorry for the delay.

PR #152, opened 2020-02-26
Extend commands support

rolang:
Hi, would be nice to have a few more commands supported :).
This adds:
- ALTER TABLE
- CREATE SCHEMA
- DROP SCHEMA


tpolecat:
Looks great! There's a problem making the test fail intermittently but I'll merge once that's fixed.

PR #174, opened 2020-04-20
gh-actions

tpolecat:


.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK
+        uses: olafurpg/setup-scala@v2
+      - name: Cache Coursier
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/coursier
+          key: sbt-coursier-cache
```
regadas, line 16: Hi @tpolecat I think you should also add an `${{ hashFiles('**/build.sbt') }}` otherwise new dependencies after this PR will always be pulled.

tpolecat: Thank you! I will add that.

image

PR #179, opened 2020-04-27
Handle ErrorResponse in Unroll.accumulate

megri:
Attempt at resolving https://github.com/tpolecat/skunk/issues/170.

I added a `case ErrorResponse(info) => syncAndFail(info)` to the accumulate-method of Unroll.scala. Generally I think it would be better to receive a single response and check that for errors before trying to accumulate data.


tpolecat:
There's a flaky test, don't worry about that. We will need a testcase for this, if you'd like to add one. If not I can.

megri:
I'll do it. Good learning opportunity :)

megri:
@tpolecat WDYT?

PR #195, opened 2020-06-12
SSL Support

tpolecat:
This adds initial SSL support. The `Session` resource methods now take an optional `ssl` named argument of type `SSL` (default is `SSL.None`). The main test database in the docker-compose file now runs with SSL enabled, with a self-signed certificate. Add `ssl = SSL.Trusted` in any test or example that uses the main test db and it should work fine.

```scala
Session.single[IO](
  host     = "localhost",
  user     = "jimmy",
  database = "world",
  password = Some("banana"),
  debug    = true,        // this now shows TLS logs as well as protocol messages
  ssl      = SSL.Trusted, // require SSL, trust all certificates
)
```

Many thanks to @mpilquist for fixing some issues in fs2's TLS support. It came together really quickly after that.

N.B. The build is on an fs2 snapshot until a new release is available.


tpolecat:
Someone needs to congratulate me on getting the #$*&#^*$ ci config to work.

PR #205, opened 2020-07-03
CommandComplete typo in decoder.

dpnova:
I'm guessing most people throw this value away. I know I use `.void` on commands heh.


tpolecat:
Thanks, this was fixed by #130 which I just got around to merging.

PR #227, opened 2020-07-21
doc odyssey

tpolecat:
- gets the transaction doc to a reasonable state
- adds channel doc
- when working on channel doc it was obvious that the parameterization was wrong, so fixed that by making `Notification` itself parameterized, and the `Channel` covariant bits delegate to that.
- also noticed that `Channel` should be a `Pipe`, and by that reasoning so should `PreparedCommand` and `PreparedQuery`, so changed those as well.
- need to add tests for this (coverage is going to drop)


tpolecat:
Oh it looks like I can sneak things past coverage by opening a draft PR and then setting it to ready. How clever of me.

tpolecat:
What happens if I close and re-open?

tpolecat:
¯\\\_(ツ)\_/¯

Issue #226, opened 2020-07-20
add overload for `.transaction` that takes parameters for `BEGIN`


tpolecat:
Right now users have to execute `SET TRANSACTION` if they desire non-default transaction characteristics. It would be nice to be able to say `s.transaction(ReadUncommitted, ReadOnly)` for instance.

PR #242, opened 2020-08-19
Add overload for `.transaction`

zsambek:
Implements #226


modules/core/src/main/scala/Session.scala
```diff
@@ -371,7 +384,10 @@ object Session {
           proto.prepare(command, typer).map(PreparedCommand.fromProto(_))
 
         override def transaction[A]: Resource[F, Transaction[F]] =
-          Transaction.fromSession(this, namer)
+          Transaction.fromSession(this, namer, ReadCommitted, ReadWrite)
```
tpolecat, line 387: I think the default here needs to map to `BEGIN` with no arguments, because the user may have done `SET SESSION CHARACTERISTICS AS TRANSACTION ...` and we want those defaults to be used, whatever they may be. So I think`Transaction.fromSession` probably wants an `Option[(TransactionIsolationLevel, TransactionAccessMode)]`.

modules/core/src/main/scala/data/TransactionAccessMode.scala
```diff
@@ -0,0 +1,21 @@
...
+
+/** Enumerated type of transaction access mode values. */
+sealed abstract class TransactionAccessMode extends Product with Serializable
+object TransactionAccessMode {
+
+  case object ReadOnly extends TransactionAccessMode
+
+  case object ReadWrite extends TransactionAccessMode
+
+  def toLiteral(accessMode: TransactionAccessMode): String =
```
tpolecat, line 15: I think a `private[skunk] sql: String` member on `TransactionAccessMode` and `TransactionIsolationLevel` might be an easier way to do this.

modules/core/src/main/scala/Session.scala
zsambek: Totally forgot about that. Fixed now :)

modules/core/src/main/scala/data/TransactionAccessMode.scala
zsambek: Looking much better, yes :)

modules/core/src/main/scala/Transaction.scala
```diff
@@ -141,11 +144,15 @@ object Transaction {
...
     def doCommit: F[Completion] =
-      s.execute(internal"COMMIT".command)
+      s.execute(internal"COMMIT".command)      
 
     val acquire: F[Transaction[F]] =
       assertIdle(CallSite("begin", Origin.unknown)) *>
-      s.execute(internal"BEGIN".command).map { _ =>
+      s.execute(
+        internal"""BEGIN
+                  ${iOp.map(i => s"ISOLATION LEVEL ${i.sql}").getOrElse("")}
```
tpolecat, line 153: Can you change this to `iOp.foldMap(i => s"ISOLATION LEVEL ${i.sql}")` here and similarly below so you don't need the `getOrElse("")`?

zsambek: Good idea! :) Done!

tpolecat (approved):

PR #247, opened 2020-08-25
applied fragments

### LINKED ISSUE #245: WHERE fragment with optional clauses
Given two fragments
```scala
  val nameLike: Fragment[String]        = sql"""name LIKE $varchar"""
  val populationLessThan: Fragment[Int] = sql"""population < $int4"""
```
is there a way to build a fragment that should behave like this one
```scala
  /**
   * Should produce:
   * 1. "" if both inputs are None
   * 2. "WHERE name LIKE $1" if first input is Some
   * 3. "WHERE population < $2" if second input is Some
   * 4. "WHERE name LIKE $1 AND population < $2" if both inputs are Some
   */
  val where: Fragment[Option[String] ~ Option[Int]] = ???
```

Related combinator from Doobie -[ `whereAndOpt`](https://github.com/tpolecat/doobie/blob/3a7297e27cb9521ad8b742944f782e7a3ee2c337/modules/core/src/main/scala/doobie/util/fragments.scala#L44)

tpolecat:
resolves #245 

@minedeljkovic how does this look? see the example

### DIFF
--- a/modules/core/src/main/scala/AppliedFragment.scala
+++ b/modules/core/src/main/scala/AppliedFragment.scala
@@ -29,6 +29,7 @@ sealed trait AppliedFragment { outer =>
 
 object AppliedFragment {
 
+  /** Construct an `AppliedFragment` from a `Fragment` and its argument. */
   def apply[A](fragment: Fragment[A], argument: A): AppliedFragment = {
     // dum dee dum
     type A0 = A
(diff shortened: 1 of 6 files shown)


minedeljkovic:
This works great!
As a general solution, it also solves another problem that I didn't mention - conditionaly setting columns in an UPDATE statement.

PR #248, opened 2020-08-25
Support all boolean styles

### LINKED ISSUE #215: support all boolean styles
The `bool` encoder should allow you to specify the true and false values you wish to use when writing. 
See https://www.postgresql.org/docs/9.6/datatype-boolean.html

zsambek:
Fixes #215


tpolecat:
This looks great but after thinking about it a little bit #248 doesn't really make any sense since all it changes is the value going over the wire. It doesn't do anything visible from the user's point of view. I'm going to close this and the related issue, it was just a mistake on my part.

PR #251, opened 2020-09-06
Correct usage of param type in implicit not found msg for Twiddler

sbuzzard:
The Twiddler type represents a mapping from an `HList` to an associated case class type with the same structure, and thus in the inductive implicit resolutions the 'A' type parameter and 'Out' type of `Twiddler` both represent `HList`s at given points, in the the evidence implicit parameter referenced in the `gmap/gcontramap/gimap` functions, the `Twiddler.Aux` is used of the form `Twitter.Aux[Target, Source]` (e.g., `ev: Twiddler.Aux[B, A]`) and thus the 'A' type parameter in `Twiddler` represents the Target from the point of view of the user of this API, so the implicit not found message should probably reflect this.  I noticed this when I had my source twiddler list structure wrong and first saw the message.   Of course this may not always be the case depending on when the resolution bails on a failure?  In which case the message could be made more ambiguous in terms of the role of 'A'?

### DIFF
--- a/modules/core/src/main/scala/util/Twiddler.scala
+++ b/modules/core/src/main/scala/util/Twiddler.scala
@@ -14,7 +14,7 @@ import scala.annotation.implicitNotFound
  * Witness that type `A` is isomorphic to a left-associated HList formed from pairs; i.e.,
  * A :: B :: C :: D :: HNil ~ (((A, B), C), D)
  */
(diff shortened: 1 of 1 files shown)


tpolecat (approved):
Great catch, thanks!

sbuzzard:
> Great catch, thanks!

thanks!  and thanks much for the lib - it's great!

PR #260, opened 2020-09-13
Removed custom SCRAM implementation in favor of library provided by...

mpilquist:
…Ongres and used in official JDBC driver. No need to maintain the SCRAM crypto stuff this way and guaranteed we won't have any weird compatibility issues.

BTW, the use of `SecureRandom.getInstanceStrong()` in the original implementation led to some hanging issues on machines without enough entropy. See https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/ for details on this, but this led me to investigate what the official JDBC client does for nonce creation. That then led me to this SCRAM library from Ongres which is nicely decoupled from everything JDBC.


mpilquist:
Any chance for a release with this PR soon?

tpolecat:
tagged v0.0.21

PR #264, opened 2020-09-14
Tests for Query

zsambek:


tpolecat:
Curiously this doesn't change test coverage. Still helpful though!

PR #269, opened 2020-09-18
remove ffstest, add munit

tpolecat:
This switches to munit, with the intent of minimizing the diff. We can go back and continue cleanup but I wanted to get this in as cleanly as possible. /cc @zsambek


tpolecat:
I'm not proud of the `Channel` test but there's no obvious way to avoid the race condition.

tpolecat:
@zsambek what do you think?

zsambek:
Looking good to me.

PR #278, opened 2020-10-02
Migrate to new kind-projector syntax

m-sp:
See also https://github.com/typelevel/kind-projector/issues/108

Btw. thanks for the project, it's a joy to use!


tpolecat:
Thanks!

Issue #229, opened 2020-07-30
Use with Redshift


bbstilson:
```
🔥  Postgres FATAL 42704 raised in set_config_option (/home/ec2-user/padb/src/pg/src/backend/utils/misc/guc.c:23873)
🔥  
🔥    Problem: Unrecognized configuration parameter "IntervalStyle".
```

IntervalStyle isn't supported by Redshift. I found some documentation that states the default parameters are not overridable at the moment, but I'd be happy to take a crack at this with some guidance in the right direction.

For example, this is what a postgres library for Julia offers: https://github.com/invenia/LibPQ.jl/blob/dc200995d2c5c928952dfe7b80d7aa70a8421956/docs/src/pages/faq.md

tpolecat:
Yep, we can provide a way to do this if you'd like to give it a go. 

The startup parameters are defined in `StartupMessage.scala`, which is referenced as a constant in the message encoder. We could instead add a `Map[String, String]` member to `StartupMessage` and encode this instead. The refactor will just be piping the parameter all the way out to the `Session` factory methods, at which point you can default it to `StartupMessage.ConnectionProperties` ... which you should move onto `Session` and call it `DefaultConnectionProperties` (be sure to update `Sessions.md` in the `docs` project, which references the constant). This would give you some wiggle room if you want to pass something different to Redshift.

Beyond that I can't really commit to supporting Redshift because there's no obvious way for me to run tests against it, but this change is minor enough and it should unblock you so you can explore a bit more.

nbenns:
I see there is a working solution in the PR but it hasn't been updated since Oct 12th, any thoughts of when it might get merged?
willing to put in some effort if need be.

tpolecat:
I was just looking at this actually. I need to resolve some conflicts but it should be merged later today.

PR #282, opened 2020-10-12
Overridable connection parameters

bbstilson:
Fixes https://github.com/tpolecat/skunk/issues/229

I found a [redshift docker container](https://github.com/GuildEducationInc/docker-amazon-redshift), which makes no promises to fully implement all the features of Redshift, but it did reproduce the error I experienced as reported in #229.

The fix involves piping connection properties through various layers as you specified (thanks!).

### DIFF
--- a/modules/core/src/main/scala/Session.scala
+++ b/modules/core/src/main/scala/Session.scala
@@ -19,7 +19,6 @@ import skunk.net.Protocol
 import skunk.util._
 import skunk.util.Typer.Strategy.{ BuiltinsOnly, SearchPath }
 import skunk.net.SSLNegotiation
-import skunk.net.message.StartupMessage
 import skunk.data.TransactionIsolationLevel
 import skunk.data.TransactionAccessMode
 
@@ -187,6 +186,13 @@ trait Session[F[_]] {
 
 /** @group Companions */
 object Session {
+  val DefaultConnectionParameters: Map[String, String] =
+    Map(
+      "client_min_messages" -> "WARNING",
+      "DateStyle"           -> "ISO, MDY",
+      "IntervalStyle"       -> "iso_8601",
+      "client_encoding"     -> "UTF8",
+    )
 
   object Recyclers {
 
@@ -253,7 +259,7 @@ object Session {
     writeTimeout: FiniteDuration = 5.seconds,
     strategy:     Typer.Strategy = Typer.Strategy.BuiltinsOnly,
     ssl:          SSL            = SSL.None,
-    parameters: Map[String, String] = StartupMessage.DefaultConnectionParameters
+    parameters: Map[String, String] = Session.DefaultConnectionParameters
   ): Resource[F, Resource[F, Session[F]]] = {
 
     def session(socketGroup:  SocketGroup, sslOp: Option[SSLNegotiation.Options[F]]): Resource[F, Session[F]] =
@@ -287,7 +293,7 @@ object Session {
     writeTimeout: FiniteDuration = 5.seconds,
     strategy:     Typer.Strategy = Typer.Strategy.BuiltinsOnly,
     ssl:          SSL            = SSL.None,
-    parameters: Map[String, String] = StartupMessage.DefaultConnectionParameters
+    parameters: Map[String, String] = Session.DefaultConnectionParameters
   ): Resource[F, Session[F]] =
     pooled(
       host         = host,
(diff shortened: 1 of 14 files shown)


modules/core/src/main/scala/Session.scala
```diff
@@ -252,10 +253,11 @@ object Session {
     writeTimeout: FiniteDuration = 5.seconds,
     strategy:     Typer.Strategy = Typer.Strategy.BuiltinsOnly,
     ssl:          SSL            = SSL.None,
+    connProps: Map[String, String] = StartupMessage.DefaultConnectionParameters
   ): Resource[F, Resource[F, Session[F]]] = {
 
     def session(socketGroup:  SocketGroup, sslOp: Option[SSLNegotiation.Options[F]]): Resource[F, Session[F]] =
-      fromSocketGroup[F](socketGroup, host, port, user, database, password, debug, readTimeout, writeTimeout, strategy, sslOp)
+      fromSocketGroup[F](socketGroup, host, port, user, database, password, debug, readTimeout, writeTimeout, strategy, sslOp, connProps)
```
tpolecat, line 260: Let's rename `connProps` to `parameters` everywhere. Postgres is very consistent about this terminology so we should use it as well.

modules/core/src/main/scala/net/message/StartupMessage.scala
```diff
@@ -19,16 +25,15 @@ case class StartupMessage(user: String, database: String) extends UntaggedFronte
 }
 
 object StartupMessage {
-
-  val ConnectionProperties: List[(String, String)] =
-    List(
+  val DefaultConnectionParameters: Map[String, String] =
```
tpolecat, line 28: Now that this is part of the public API let's move `DefaultConnectionParameters` to the `Session` companion so users don't need to dig into the protocol implementation to find it.

modules/docs/src/main/paradox/reference/Sessions.md
```diff
@@ -54,14 +54,26 @@ In addition to these options, the `SSL` values themselves allow the following mo
...
+You may set arbitrary session parameters via the `connProps` property in your session initialization. For example, to use with Amazon Redshift:
+
+```scala mdoc:compile-only
+Session.single[IO](
+  host     = "localhost",
+  user     = "jimmy",
+  database = "world",
+  password = Some("banana"),
+  port = 5439,
+  connProps = StartupMessage.DefaultConnectionParameters
```
tpolecat, line 76: `DefaultConnectionParameters - "IntervalStyle"` should work ... ?

modules/docs/src/main/paradox/reference/Sessions.md
```diff
@@ -54,14 +54,26 @@ In addition to these options, the `SSL` values themselves allow the following mo
...
 | Parameter | Value |
 |----------|-------|
 ```scala mdoc:passthrough
-println(StartupMessage.ConnectionProperties.map { case (k, v) => s"| `$k` | `$v` |" } .mkString("\n"))
+println(StartupMessage.DefaultConnectionParameters.map { case (k, v) => s"| `$k` | `$v` |" } .mkString("\n"))
 ```
 
-Future versions of Skunk may be more flexible in this regard, but for now your application needs to be ok with these defaults.
\ No newline at end of file
+You may set arbitrary session parameters via the `connProps` property in your session initialization. For example, to use with Amazon Redshift:
```
tpolecat, line 67: We need to make it clear that this is unsupported in general. I suggest wording like

> It is possible to modify default session parameters via the `parameters` session property, which is unsupported in general but may be necessary when using nonstandard Postgres variants. Amazon Redshift for example does not support the `IntervalStyle` parameter, and this will cause startup negotiation to fail. A workaround is demonstrated below.

modules/tests/src/test/scala/RedshiftTest.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  test("redshift - successfully connect") {
+    Session.single[IO](
+      host = "localhost",
+      user = "postgres",
+      database = "postgres",
+      password = None,
+      port = 5439, // redshift port
+      connProps = StartupMessage.DefaultConnectionParameters
+        .filterNot { case (k, _) => k == "IntervalStyle" }
```
tpolecat, line 23: `map - key`?

tpolecat (changes requested):
This looks great. I'd like to make a few superficial changes.

modules/docs/src/main/paradox/reference/Sessions.md
bbstilson: I so rarely delete keys from maps I completely forgot haha. Thanks.

bbstilson:
Thanks for the feedback @tpolecat. I believe I addressed everything, but let me know if I missed anything.

tpolecat:
subsumed by #362

PR #290, opened 2020-11-02
Make tutorial more Scala-newbie friendly

visortelle:
It takes about 15mins for me to determine the reason for the following error:

```
Could not find an instance of Concurrent for cats.effect.IO.
I found:

    cats.effect.IO.ioConcurrentEffect(
      /* missing */summon[cats.effect.ContextShift[cats.effect.IO]]
    )

But no implicit values were found that match type cats.effect.ContextShift[cats.effect.IO].

One of the following imports might make progress towards fixing the problem:

  import shapeless.~?>.idKeyWitness
  import shapeless.~?>.idValueWitness
  import shapeless.~?>.witness


  )
```

Thanks for your work! 👍

### DIFF
--- a/modules/docs/src/main/paradox/tutorial/Setup.md
+++ b/modules/docs/src/main/paradox/tutorial/Setup.md
@@ -32,8 +32,10 @@ import skunk._
 import skunk.implicits._
 import skunk.codec.all._
 import natchez.Trace.Implicits.noop                          // (1)
+import scala.concurrent.ExecutionContext
 
 object Hello extends IOApp {
+  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
 
   val session: Resource[IO, Session[IO]] =
     Session.single(                                          // (2)


tpolecat:
What is reporting that error message?

That instance is provided automatically by `IOApp`, you don't need to declare it.

visortelle:
The error happens on running the following test:

```
package xxx.server.postgresql.testing

import org.scalatest._
import org.scalatest.matchers.should.Matchers
import org.scalatest.flatspec.AnyFlatSpec
import cats.effect._
import skunk._
import skunk.implicits._
import skunk.codec.all._
import natchez.Trace.Implicits.noop
import scala.concurrent.ExecutionContext

class TestContainerTest extends AnyFlatSpec with Matchers {
  behavior of "TestContainer"

  val pg = PostgreSQLInstance()
  pg.start()
  val pgConfig = pg.getConfig

  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)

  val session: Resource[IO, Session[IO]] = Session.single(
    host = pgConfig.host,
    port = pgConfig.port.toInt,
    database = pgConfig.database,
    user = pgConfig.user,
    password = Some(pgConfig.password)
  )

  it should "work like in this example" in {
    // no code
  }
}
```

I use dotty `0.27.0-RC1` maybe it can be a reason for the issue?

visortelle:
My fault. Didn't notice:

> Try out this minimal IOApp that connects to the database and selects the current date.

Closing.

tpolecat:
Also Skunk doesn't work in Scala 3 yet.

PR #292, opened 2020-11-03
Scala 3

tpolecat:
This adds a build for Scala 3.0.0-M1, which is mostly just for show at this point but I really needed to be confident the sql interpolator would be possible and yes, it works fine. I'm unhappy with `Twiddler.scala` but the weird reduction semantics of match types and https://github.com/lampepfl/dotty/issues/9871 make it impossible to do any better as far as I can tell. Eventually I'd like to get rid of twiddle lists since you can use `Tuple` in Scala 3, but as a practical matter I think that's all a long way off. Super disappointed with my experience so far.

- [x] split up scala 2/3 bits
- [x] syntactic changes
- [x] implement `Origin`
- [x] implement `sql` interpolator
- [x] implement `Twiddler`
- [x] get examples compiling in Scala 3
- [x] get tests running in Scala 3 (mostly)
- [x] get coverage running again

Also added an alternative to twiddle-lists which lets you say `ca *: cb *: cc` to get a `Codec[(A, B, C)]` and generic mapping with `pimap[ProductType]`, and analogously `pmap` and `pcontramap` with `Decoder` and `Encoder`. This is more efficient and fits the Scala-3 lifestyle better than twiddle lists. Both are supported for now.


tpolecat:
Tried a lot of things but th best I could do for `Twiddler` implementations was to hardcode them up to arity 5. We will certainly revisit but this should be sufficient to get the tests running.

tpolecat:
Have to punt on tests for now because `munit` doesn't worth with dotty compat (you need the dotty version) which is find, but I also need `cats-effect-munit` which isn't available for 3.0.0-M1 yet because _its_ upstream stuff isn't available. So we'll revisit.

PR #315, opened 2020-11-24
Decoder can have an Applicative, not just Apply

stephenjudkins:


stephenjudkins:
@tpolecat what would I need to do to get this merged? Some of the code coverage changes reported here look spurious...

tpolecat:
Ok sorry for the delay. I just tagged v0.0.22 so it should be on sonatype in the next hour or so.

PR #336, opened 2020-12-12
initial ce3 port

tpolecat:
So far so good. Issues remaining:

- [x] waiting on Natchez for CE3 (which is waiting on Log4Cats)
- [ ] a bunch of pool tests fail due to differences in error handling with bracket in CE3, see https://github.com/typelevel/cats-effect/issues/1494
- [x] fs2 SocketGroup demands Sync but should only demand Network. When this is fixed the Async requirement for session construction will weaken to Concurrent. https://github.com/typelevel/fs2/pull/2182
- [x] awaiting non-snapshot version of fs2 with the above change
- [ ] Session constructors take Console but it's only used for debug output, so it would be nice to weaken that and only allow it if the user requests debug output.
- [ ] http4s example won't work because http4s isn't available for CE3 yet. Probably won't wait on this one.
- [x] ce3 Console doesn't work with mdoc because it uses stdlib Console.withOut which doesn't affect ce3 Console. Will try to fix in mdoc.


RaasAhsan:
2 will be fixed by https://github.com/typelevel/cats-effect/pull/1549, which should be available in CE3-M6.

tpolecat:
Ah, i see what i did. I should have merged master into series/0.1. Managing multiple branches is so error-prone.

PR #362, opened 2021-01-29
override connection properties

### LINKED ISSUE #229: Use with Redshift

tpolecat:
Fixes #229
subsumes #282

### DIFF
--- a/bin/local
+++ b/bin/local
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
-set -eu
+set -e
 
 DIR=$1
+EXTRA_FLAGS=$2
 
 print_usage() {
   printf "\nUsage:\n \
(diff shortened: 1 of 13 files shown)


tpolecat:
subsumed by #362

tpolecat:
@bbstilson sorry for the delay on this, we should be good now.

/cc @nbenns

tpolecat:
released v0.0.23

PR #378, opened 2021-02-14
Fix _float4 and _float8 Codecs to work on arrays not primitive values

lorandszakacs:
Seemed like one those pesky copy + paste oversights.


tpolecat:
D'oh!

PR #399, opened 2021-03-02
CE3 (conflicts resolved)

tpolecat:
Subsumes #398 

Remaining issues (promoted to issues):

- [x] Session constructors take Console but it's only used for debug output, so it would be nice to weaken that and only allow it if the user requests debug output. #400
- [x] http4s example won't work because http4s isn't available for CE3 yet. Probably won't wait on this one. #401


m-sp:
btw there is a [milestone of http4s](https://github.com/http4s/http4s/releases/tag/v1.0.0-M18) which is published for cats-effect RC2

PR #402, opened 2021-03-02
Reenable http4s

m-sp:
fixes https://github.com/tpolecat/skunk/issues/401

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -183,9 +183,9 @@ lazy val example = project
       "org.tpolecat"  %% "natchez-honeycomb"   % natchezVersion,
       "org.tpolecat"  %% "natchez-jaeger"      % natchezVersion,
     ) ++ Seq(
-      "org.http4s"    %% "http4s-dsl"          % "1.0.0-M17",
-      "org.http4s"    %% "http4s-blaze-server" % "1.0.0-M17",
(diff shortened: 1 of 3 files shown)


tpolecat:
Hey that was fast, thanks!

m-sp:
ugh seems like a nasty circe eviction, http4s depends on the milestone already :( any advice how to fix it?

tpolecat:
Hang on I'll have a look.

tpolecat:
Kind of sucks because that circe release didn't publish for Scala 3.0.0-M3.

m-sp:
yep rather unfortunate that circe milestones only publish for one Scala version... I can revisit a bit later

tpolecat:
Yeah http4s milestone is also only out for Scala 3..0.0-RC1.

tpolecat:
There's kind of a cascade of things that will need fixing here, so I'm closing this one for now.

I think what we need to do first is get all of `master` building for Scala 3 (circe module and http4s example) and drop the Scala 3.0.0-M3 cross-build. Then once that's merged create a branch from series/0.1, merge master into it, and use that as the basis for getting it all working here.

tpolecat:
An added note here, Circe 0.14 is going to break bincompat, so we can't really force people to upgrade because other things (like http4s 0.21) are depending on 0.13. So I think we don't have a move at the moment.

PR #448, opened 2021-05-03
RETURNING docs

antonkw:


tpolecat:
Thanks! This needs to go somewhere else in the doc but I'm not sure where. I will fold it in next time I work on docs.

PR #506, opened 2021-07-08
Add notifications method to Session

dbdbdb:
The documentation state that there should be a notifications function
on the Session to get events from all subscribed channels:

https://tpolecat.github.io/skunk/tutorial/Channels.html#listening-to-a-channel

The function existed on the protocol level but was missing in
the Session.


matthughes:
I know this is pretty old but has this been overcome by events @dbdbdb ?  There is a `listen` method on `Session` that has the same type signature.  How is that different than your proposed method?

PR #515, opened 2021-07-29
Cross-build for Scala.js

armanbilge:
Blocked by https://github.com/tpolecat/SourcePos/pull/6 (cc @ChristopherDavenport) and also ~problems with scodec-bits.~ Update: I worked around the scodec issue. Just waiting on sourcepos and hopefully should be good to go? I'm have trouble running tests on local so going to be relying on CI for this one.

This is actually very nearly a pure cross. The most notable difference for JS was resurrecting the scram implementation courtesy of @mpilquist that was subsequently scrapped in https://github.com/tpolecat/skunk/pull/260 as well as adding an npm dependency to [saslprep](https://www.npmjs.com/package/saslprep).


modules/core/js/src/main/scala/net/message/Scram.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  private def H(input: ByteVector): ByteVector = input.digest("SHA-256")
+
+  private def Hi(str: String, salt: ByteVector, iterations: Int): ByteVector = {
+    // TODO It is unfortunate that we have to use a sync API here when an async is available
+    // To make the change here will require running an F[_]: Async up the hiearchy    
+    val salted = crypto.pbkdf2Sync(str, salt.toByteBuffer.arrayBuffer(), iterations, 8 * 32, "sha256")
+    bufferToByteVector(salted)
+  }
```
armanbilge, line 100: FYI.

mpilquist: Can you elaborate? PBKDF-2 should be completely CPU bound

armanbilge: Node.js offers an async [version of this method with a callback](https://nodejs.org/api/crypto.html#crypto_crypto_pbkdf2_password_salt_iterations_keylen_digest_callback). I assume they have a good reason for doing so.

My best guess would be that you are indeed right that it is CPU-bound, but it's CPU intensive enough that it would detrimentally block a single-threaded Node.js app.

tpolecat:
I appreciate your effort on this but I'm not convinced that anyone actually needs this. It's a significant complication and maintenance burden going forward. Can you justify this a little bit?

armanbilge:
No worries, completely understandable! I recently got http4s cross-building to JS (including ember-server) in https://github.com/http4s/http4s/pull/4938. Many people seem interested in using this with Node.js on AWS lambda and/or similar services. Skunk would be an important part of that I assume.

armanbilge:
Thanks to fs2 3.1.0, all tests are passing now here. This is ready, all it needs is https://github.com/tpolecat/SourcePos/pull/6.

modules/core/js/src/main/scala/net/message/Scram.scala
```diff
@@ -0,0 +1,148 @@
...
+    val hash = crypto.createHash("sha256")
+    hash.update(byteVectorToUint8Array(input))
+    bufferToByteVector(hash.digest())
+  }
+
+  private def Hi(str: String, salt: ByteVector, iterations: Int): ByteVector = {
+    // TODO It is unfortunate that we have to use a sync API here when an async is available
+    // To make the change here will require running an F[_]: Async up the hiearchy    
+    val salted = crypto.pbkdf2Sync(str, byteVectorToUint8Array(salt), iterations, 8 * 32, "sha256")
+    bufferToByteVector(salted).take(32)
```
armanbilge, line 111: @mpilquist you wouldn't happen to have any guesses why this `take(32)` is necessary here, but not on JVM?

tpolecat:
Amazing! I'm out for the day but will have a closer look later this weekend.

tpolecat:
Ok SourcePos 1.0.1 has been published. It may take a little while to appear on Central.

modules/core/js/src/main/scala/SSLPlatform.scala
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2018-2021 by Rob Norris
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package skunk
+
+private[skunk] trait SSLCompanionPlatform {
+  
+}
```
tpolecat, line 9: Braces are unnecessary.

modules/core/jvm/src/main/scala/SSLPlatform.scala
```diff
@@ -11,60 +11,15 @@ import java.security.KeyStore
...
-
-  /** `SSL` which indicates that SSL is not to be used. */
-  object None extends SSL() {
-    def tlsContext[F[_]: Network](implicit ev: ApplicativeError[F, Throwable]): F[TLSContext[F]] =
-      ev.raiseError(new Exception("SSL.None: cannot create a TLSContext."))
-    override def withFallback(fallbackOk: Boolean): SSL = this
-    override def withTLSParameters(tlsParameters: TLSParameters): SSL = this
-  }
 
+private[skunk] trait SSLCompanionPlatform { self: SSL.type =>
```
tpolecat, line 15: If you're not using the `self` alias somewhere you can use `this: SSL.type =>` instead.

modules/core/shared/src/main/scala/SSL.scala
```diff
@@ -0,0 +1,59 @@
...
+      case SSL.None => none.pure[F]
+      case _ => tlsContext.map(SSLNegotiation.Options(_, tlsParameters, fallbackOk, logger).some)
+    }
+
+}
+
+object SSL extends SSLCompanionPlatform {
+
+  /** `SSL` which indicates that SSL is not to be used. */
+  object None extends SSL() {
```
tpolecat, line 46: You don't need the parens in `SSL()` here or below.

modules/tests/jvm/src/test/scala/SslTestPlatform.scala
```diff
@@ -0,0 +1,43 @@
...
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package tests
+
+import cats.effect._
+import skunk._
+import natchez.Trace.Implicits.noop
+
+trait SslTestPlatform { self: SslTest =>
```
tpolecat, line 11: Why are these tests not run for JS?

tpolecat (commented):
Some very minor things, but also can you explain why scala-js isn't running all the SSL tests?

modules/core/shared/src/main/scala/SSL.scala
armanbilge: Yup, just copying your style :P

modules/tests/jvm/src/test/scala/SslTestPlatform.scala
tpolecat: So you just have to tell node to ignore EVERY certificate? You can't do it per connection? That's kind of nuts.

armanbilge: I think you can do it on the _global_ level with the environment variable and possibly on the _socket_ level by fiddling with the TLS parameters: https://github.com/typelevel/fs2/blob/main/io/js/src/main/scala/fs2/io/net/tls/TLSParameters.scala#L44

But, no way to do it for an entire context.

armanbilge:
Should I be concerned and/or do something about the codecov/patch failure?

tpolecat:
No, don't worry about coverage.

tpolecat:
Any last words before I slam the button?

armanbilge:
Lol nope, thanks for merging!!

PR #529, opened 2021-08-18
Set TCP_NODELAY

### LINKED ISSUE #450: Option to set TCP_NODELAY
Recently, I have noticed my query preparations and executions take ~50ms but I could not tell why.
Today, I got the luck to find [this](https://mysteries.wizardzines.com/50ms-request.html).
I checked and this is the issue I have with skunk and postgres.

To lower my query latencies, I would like to set `TCP_NODELAY` in skunk. The option is supported in fs2 `Socket` and it should only be a matter of making it available in `Session.pooled`.

trobert:
We have faced the same issue as described here: https://github.com/tpolecat/skunk/issues/450
I propose to set this option, it avoids a 40ms penalty in each round trip to the server.

btw, would you also take an identical PR for the 0.0.x branch ?
Thanks

fixes #450

### DIFF
--- a/modules/core/shared/src/main/scala/net/BitVectorSocket.scala
+++ b/modules/core/shared/src/main/scala/net/BitVectorSocket.scala
@@ -9,7 +9,7 @@ import cats.effect._
 import cats.syntax.all._
 import fs2.Chunk
 import scodec.bits.BitVector
-import fs2.io.net.{ Socket, SocketGroup }
+import fs2.io.net.{Socket, SocketGroup, SocketOption}
 import com.comcast.ip4s._
 import skunk.exception.EofException
 
@@ -67,7 +67,7 @@ object BitVectorSocket {
     sslOptions:   Option[SSLNegotiation.Options[F]],
   )(implicit ev: MonadError[F, Throwable]): Resource[F, BitVectorSocket[F]] =
     for {
-      sock  <- sg.client(SocketAddress(Hostname.fromString(host).get, Port.fromInt(port).get)) // TODO
+      sock  <- sg.client(SocketAddress(Hostname.fromString(host).get, Port.fromInt(port).get), List(SocketOption.noDelay(true))) // TODO
       sockʹ <- sslOptions.fold(sock.pure[Resource[F, *]])(SSLNegotiation.negotiateSSL(sock, _))
     } yield fromSocket(sockʹ)
 


tpolecat:
Thanks. Are there any times that you would _not_ want this option? i.e., should it be configurable?

tpolecat:
And yes, assuming this works out I can take a PR for 0.0.x, but only if you know of someone who needs it. I think most users have upgraded to CE3.

trobert:
I don't know whether some people could not want this. But in our case at least, it just seems to be  a much better default.

I quote the details from https://mysteries.wizardzines.com/50ms-request.html because I can't make a direct link to the answer:

> This bug seems obscure if you haven't seen it before, but it's relatively common! It can happen any time a client is sending two small packets immediately followed by reading a response.Here are 5 blog posts talking about the same bug:
> 
> [40 milliseconds of latency that just would not go away](https://rachelbythebay.com/w/2020/10/14/lag/)
> [Why you should understand (a little) about TCP](https://jvns.ca/blog/2015/11/21/why-you-should-understand-a-little-about-tcp/)
> [In search of performance - how we shaved 200ms off every POST request](https://gocardless.com/blog/in-search-of-performance-how-we-shaved-200ms-off-every-post-request/)
> [40 millisecond bug](https://vorner.github.io/2020/11/06/40-ms-bug.html)
> [TCP Performance problems caused by interaction between Nagle’s Algorithm and Delayed ACK](http://www.stuartcheshire.org/papers/NagleDelayedAck/)
> A "magic number" for this bug is 40 milliseconds — that's the delayed ACK timeout. But like you can see in the "how we shaved 200ms off every POST request" blog post above, it can also manifest as a 200ms delay in some cases.
> 
> Because of this issue, a lot of programming languages (including Python, Ruby, and Go) set TCP_NODELAY on every TCP connection by default to disable Nagle's algorithm.
> 
> Also, the Linux kernel doesn't always enable delayed ACKs — to reproduce this I actually had to write a Python program that explicitly turns them on. I haven't been able to find a clear explanation of exactly when delayed ACKs are used.

AFAIU there is no pipelining in skunk yet, so this makes this issue much more visible. 

On a local postgres, here the numbers I get from tracing a few requests:
```
without TCP_NODELAY (ms)
pool.allocate	0
parse	43
parse	2
parse	1
parse	43
query	2
bind	2
execute	25
close-portal	42
bind	1
execute	1
close-portal	41
query	1
close-statement	1
close-statement	42
close-statement	1
close-statement	1
pool.free	3

with TCP_NODELAY (ms)
pool.allocate	0
parse	2
parse	1
parse	1
parse	1
query	2
bind	2
execute	33
close-portal	2
bind	1
execute	1
close-portal	1
query	1
close-statement	1
close-statement	1
close-statement	1
close-statement	1
pool.free	2

```

tpolecat:
Alright I'm convinced. Thanks!

PR #534, opened 2021-08-27
Add PostGIS support.

cranst0n:
Preliminary crack at adding PostGIS codecs, using the PostGIS JDBC driver types. I based of this stuff on the way PostGIS support is done in doobie.

Note that this changes the Docker image to add the PostGIS extension, which may have CI consequences I can't foresee, but it works on my machine :sunglasses:.


build.sbt
```diff
@@ -159,10 +159,20 @@ lazy val circe = crossProject(JVMPlatform, JSPlatform)
     )
   )
 
+lazy val postgis = crossProject(JVMPlatform, JSPlatform)
+  .crossType(CrossType.Pure)
```
armanbilge, line 163: Yay!!

cranst0n:
@tpolecat I've reworked it and wanted to get an initial opinion from you if this direction seems okay to you. I'm jumping in the deep on PostGIS. Seems like being able to decode geometries from Well Known Binary (WKB) and encode them using Well Known Text may be a path forward.

tpolecat:
I will try to have a look at this soon. Thanks!

cranst0n:
So I'm almost apologetic about this, but the scope of this expanded far more than I originally intended. Some of this probably isn't necessary to take on in skunk and could be broken into its own library, but ultimately made it easier to create tests to build my own confidence. Plus is gave me an excuse to finally play around with cats-parse.

Definitely looking for feedback, the good, the bad and the ugly. Willing to massage this into a form you're comfortable with. Or even blow it up and take it back to basics.

modules/postgis/src/main/scala/ewkb/codecs.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  def ewkbType(implicit byteOrdering: ByteOrdering): Scodec[EWKBType] =
+    guint32.xmap(EWKBType.fromRaw, EWKBType.toRaw)
+
+  def srid(implicit byteOrderering: ByteOrdering, ewkb: EWKBType): Scodec[Option[SRID]] = {
+    optional(provide(ewkb.srid == EWKBSRID.Present), gint32.xmap(SRID.apply, _.value))
+  }
+
+  def coordinate(implicit byteOrdering: ByteOrdering, ewkb: EWKBType): Scodec[Coordinate] = {
+    // TODO: Why is this val assignment necessary? Scala 3 complains without it:
```
cranst0n, line 86: FYI, this issue seems to go away  when using scala 3.0.2 instead of 3.0.1

k0ala:
Very interesting! Is this PR still under consideration?

mpilquist:
If someone wants to get this in to a mergeable state again, for sure.

mpilquist:
Thanks @cranst0n!

PR #541, opened 2021-09-04
JS byte buffer cleanups

armanbilge:
Brings in the latest scodec-bits with better interop for JS, enabling some cleanups!


armanbilge:
I forgot about this one, would be good to merge as well.

PR #544, opened 2021-09-09
Add codec for LTREE type.

cranst0n:
Adds a codec for the postgresql [ltree](https://www.postgresql.org/docs/current/ltree.html) type.


tpolecat:
Cool! I will have a look.

modules/core/shared/src/main/scala/data/LTree.scala
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) 2018-2021 by Rob Norris
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package skunk.data
+
+import cats.Eq
+
+final case class LTree private (labels: List[String]) {
```
tpolecat, line 9: This needs to be `sealed abstract case class` in order to suppress `copy` and companion `apply`.

modules/core/shared/src/main/scala/data/LTree.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  def isAncestorOf(other: LTree): Boolean = {
+    labels.isEmpty ||                 // Empty LTree is parent to all
+    other.labels.startsWith(labels)   // Other labels starts with this labels
```
tpolecat, line 13: All lists start with the empty list so the `.isEmpty` case isn't necessary.

modules/core/shared/src/main/scala/data/LTree.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  def isDescendantOf(other: LTree): Boolean = other.isAncestorOf(this)
+  
+  override def toString: String = labels.mkString(LTree.Separator.toString())
+}
+
+object LTree {
+  val Empty = LTree(Nil)
+
+  def unsafe(labels: String*): LTree = LTree(labels.toList)
```
tpolecat, line 24: Let's leave this out. The user can deal with the case where they're sure it's valid. We could also add a `ltree"..."` literal handler at some point.

modules/core/shared/src/main/scala/data/LTree.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  def fromString(s: String): Either[String, LTree] = {
```
tpolecat, line 26: It seems like we could do all of this with a single regex.

modules/core/shared/src/main/scala/data/LTree.scala
```diff
@@ -0,0 +1,60 @@
...
+      else
+        Right(LTree(labels))
+    }
+  }
+
+  val MaxLabelLength = 255
+  val MaxTreeLength = 65535
+  
+  private val Separator = '.'
+  private val ValidLabelRegex = s"""^[A-Za-z0-9_]{1,$MaxLabelLength}$$""".r
```
tpolecat, line 57: This is actually more restrictive than the implementation, which apparently depends on the character encoding, which Skunk requires to be UTF-8. So `foo.βar.baz` is a legal `lpath`. But unless we look at the source it's unlikely we'll be able to figure out exactly what it's looking for.

tpolecat (commented):
A few comments. We also need to add it to the schema types doc page.

modules/core/shared/src/main/scala/data/LTree.scala
cranst0n: Thought this issue was fixed in `2.13.something` but good point.

modules/core/shared/src/main/scala/data/LTree.scala
cranst0n: I adjusted the label regex such that it seems to pass your example but I'm sure that doesn't necessarily meet the spec.

modules/core/shared/src/main/scala/data/LTree.scala
cranst0n: You're probably right. My regex-foo is limited but I can try to tackle it when I get a little more free time.

modules/core/shared/src/main/scala/data/LTree.scala
cranst0n: I can try to add a literal. I've never built one before but seems like the `Literally` library might be a good way to go? Or would you prefer to avoid adding the dependency? At a quick glance it looks like it would need to be done independently for scala 2/3.

modules/core/shared/src/main/scala/data/LTree.scala
tpolecat: Yeah we still build for 2.12

modules/core/shared/src/main/scala/data/LTree.scala
tpolecat: Yes, Literally is the way to go. We can save it for a followup.

modules/core/shared/src/main/scala/data/LTree.scala
tpolecat: Yeah the spec is probably "whatever the implementation does", which we might want to look at. It would be nice to follow the same rules on the Skunk end.

cranst0n: I think [this](https://github.com/postgres/postgres/blob/c30f54ad732ca5c8762bb68bbe0f51de9137dd72/src/backend/regex/regc_pg_locale.c#L332) is the implementation used to check is the character is allowable. My C is more than rusty but I'll have a closer look this weekend.

cranst0n: Also, looks like my updated regex may have broken the tests on the ScalaJS side.

ohuu:
Is this PR ready to merge? I'm using ltree in a project

PR #546, opened 2021-09-10
Update fs2 to 3.1.4

armanbilge:
Supersedes #545. This updates the `yarn.lock` files which steward doesn't know about.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ lazy val `scala-2.13` = "2.13.6"
 lazy val `scala-3.0`  = "3.0.2"
 
 // This is used in a couple places
-lazy val fs2Version = "3.1.1"
+lazy val fs2Version = "3.1.2"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -185,7 +185,8 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform)
   )
   .jsSettings(
     Test / scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule)),
-    useYarn := true
+    useYarn := true,
+    yarnExtraArgs += "--frozen-lockfile",
```
armanbilge, line 189: I forgot to add this in the original PR. `--frozen-lockfile` prevents the lockfiles from changing by themselves during CI, which would be bad.

armanbilge:
Npm yarn whatever gone forever!

Issue #524, opened 2021-08-12
Broken Pipe Exception from Pool


channingwalton:
I see this exception that seems related to Skunk, which happens randomly and not even when there is any interaction with the DB.

There was a suggestion it is a connection loss which I'm investigating.

```
java.io.IOException: Broken pipe
at apply @ skunk.util.Pool$.poolImpl$1(Pool.scala:80)
at java.base/sun.nio.ch.EPollPort$EventHandlerTask.run(EPollPort.java:306)
at flatMap @ fs2.io.net.SocketCompanionPlatform$AsyncSocket.go$2(SocketPlatform.scala:132)
at java.base/sun.nio.ch.UnixAsynchronousSocketChannelImpl.onEvent(UnixAsynchronousSocketChannelImpl.java:213)
at acquireN @ fs2.Stream.waitN$1(Stream.scala:1392)
at java.base/sun.nio.ch.AsynchronousChannelGroupImpl$1.run(AsynchronousChannelGroupImpl.java:112)
at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
at java.base/sun.nio.ch.FileDispatcherImpl.write0(Native Method)
at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:50)
at java.base/sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:47)
at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
at java.base/sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:113)
at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:79)
at java.base/java.lang.Thread.run(Thread.java:829)
at async_ @ fs2.io.net.SocketCompanionPlatform$AsyncSocket.go$2(SocketPlatform.scala:126)
```

sebastianvoss:
@channingwalton I'm currently seeing similar errors. Did you figure out the root cause?

channingwalton:
> @channingwalton I'm currently seeing similar errors. Did you figure out the root cause?

Hi, no unfortunately I didn’t have time. We’ve switched to ember which doesn’t suffer the same issue.

ahjohannessen:
[...]

PR #558, opened 2021-10-04
Session: add List[SocketOption] as param

agustafson:
We're having issues similar to #524, and thought that setting the keep-alive socket option might help to resolve that issue. This PR exposes the socket options as a param in `Session`.


agustafson:
FYI tested it out and adding `SocketOption.keepAlive(true)` didn't seem to solve the underlying `BrokenPipe` issue. Think this PR might still be useful.

hnaderi:
@agustafson Can you describe the situation which this issue occurs? does it happen locally or on deployment?
Do you maintain long lived connections?
Is your deployment on docker swarm, k8s , ...?

mpilquist:
Thanks!

mpilquist:
Thanks!

Issue #576, opened 2021-11-17
Refined module is currently published as org.tpolecat:refined


LukaJCB:
I assume it's supposed to be `skunk-refined`? Should be as simple as adding `name := "skunk-refined"` to the module definition.

tpolecat:
Doh! Thank you!

PR #581, opened 2021-11-29
Flesh out Refined module a bit

LukaJCB:
Looking for some early feedback on this one. Let me know if you have any thoughts on code structure and the like. 
I removed the NumericCodec file since it looked like a WIP kind of thing.


tpolecat:
This looks great to me. I like the `.refined` syntax.

LukaJCB:
Looks like for some reason `refined-cats` is not published for 3.x, gonna follow up on their side.

LukaJCB:
Also pushed a fix for #576 :)

LukaJCB:
Looks like we ran into the infamous Scala 3.1.x and 3.0.x incompatability 😬

tpolecat:
> Looks like we ran into the infamous Scala 3.1.x and 3.0.x incompatability 😬

Hooray! Is Refined being published on 3.1?

fthomas:
Yes, it is.

tpolecat:
@fthomas could Refined be published with 3.0, or does it need stuff introduced in 3.1? I'm not super opposed to upgrading Skunk but I don't want to do it unless I have to.

fthomas:
I guess it could be published with 3.0 but hasn't the 3.0 ship sailed already? FS2 >= 3.2 is on Scala 3.1 and I guess sooner or later you want to use a newer FS2 version here too.

tpolecat:
Ok well if FS2 is on 3.1 I guess we can upgrade. Lucky for us nobody is actually using Scala 3 :-P

LukaJCB:
Updated to Scala 3.1.x and FS2 3.2.x

LukaJCB:
@tpolecat let me know how you'd like to proceed with this one :)

tpolecat:
Yes, definitely! Mark it as ready for review when you think it's in good shape.

lenguyenthanh:
Hey, I'm using this code with skunk 0.6.0, and it works great. So, I'm just wondering do we still want this? If yes, I can work on this by cherry pick the commits, and do any further works that are required.

PR #587, opened 2021-12-11
Switch to Scala.js dependency for saslprep

armanbilge:
This replaces the npm dependency for saslprep with a pure Scala.js implementation. This lets us to dump npm, dump ScalaJSBundler, dump the 8000 lines of lockfiles, and all the associated maintenance burden.

The Scala saslprep implementation is just a port from [MongoDB](https://github.com/mongodb/mongosql-auth-java/blob/master/src/main/java/org/mongodb/mongosql/auth/plugin/SaslPrep.java). I originally started adding this code directly to skunk, but I ended microlib-ing it instead because it seemed to be getting out-of-scope. It's a [single file](https://github.com/armanbilge/saslprep/blob/main/src/main/scala/com/armanbilge/SaslPrep.scala) and test suite so happy to just include it in skunk if you don't mind the bloat (i.e. a stupid number of hardcoded UTF8 codepoints. Frustrating that Node.js doesn't have this stuff out-of-the-box).


tpolecat:
BRILLIANT!

keynmol:
> This lets us to dump npm, dump ScalaJSBundler, dump the 8000 lines of lockfiles, and all the associated maintenance burden

Must say, this entire sentence made me all hot and bothered. Very arousing.

PR #599, opened 2021-12-28
Implement AuthenticationCleartextPassword

astiob:
Hit this in a real project.

NB: I’ve done my best to compare this to the docs and [PgJDBC](https://github.com/pgjdbc/pgjdbc/blob/06cecbb00f4babb455303d0330e995095f70d59d/pgjdbc/src/main/java/org/postgresql/core/v3/ConnectionFactoryImpl.java#L675-L695), but I must admit I have not tested this (yet) beyond the fact that it compiles. Ideally this would have a test right within the Skunk repo, but I’m not sure how to do it.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,6 +33,17 @@ services:
       POSTGRES_PASSWORD: banana
       POSTGRES_HOST_AUTH_METHOD: scram-sha-256
       POSTGRES_INITDB_ARGS: --auth-host=scram-sha-256
+  # for testing clear-text password login
+  password:
+    image: postgres:11
+    ports:
+      - 5435:5432
+    environment:
+      POSTGRES_DB: world
+      POSTGRES_USER: jimmy
+      POSTGRES_PASSWORD: banana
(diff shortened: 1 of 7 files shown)


tpolecat:
Oh cool, thanks. To test this we need to add another instance to the docker-compose setup that's configured to use cleartext passwords. Also need to update doc. I'm going to switch this to a draft PR until those things are done. If you'd like to give it a shot, look at the authentication tests and give me a heads-up. Otherwise I'll get to it at some point.

tpolecat:
This looks great, thanks!

tpolecat:
Please punch "ready for review" when you're happy with it.

astiob:
I’ve re-pushed with documentation fixes (somehow I had missed the fact this was documented at all, sorry!) and with a tests in a separate commit, based on the existing ones.

I saw the earlier CI run fail, but I think it shouldn’t have been caused by the auth changes.

astiob:
Aha, now I think it failed because it was specifically checking that cleartext wasn’t supported. I’ve removed that and re-pushed.

PR #611, opened 2022-02-01
Add note about IDE setup

Daenyth:


modules/docs/src/main/paradox/tutorial/Setup.md
```diff
@@ -22,6 +22,12 @@ Create a new project with Skunk as a dependency.
   version="$version$"
 }
 
+## IDE Setup
+
+For metals, no additional steps are needed.
+
+For IntelliJ, add the [skunk-intellij plugin](https://github.com/trobert/skunk-intellij) to have correct code highlighting support for queries
```
tpolecat, line 29: Lst's s/queries/statements and add a period at the end ;-)

Thanks!

PR #617, opened 2022-02-05
Host and port validation for BitVectorSocket

### LINKED ISSUE #616: need real errors for bad hostname/port
If you use an invalid hostname or port Skunk will crash with `None.get` [here](https://github.com/tpolecat/skunk/blob/main/modules/core/shared/src/main/scala/net/BitVectorSocket.scala#L70). We need to raise useful messages here.

zsambek:
Fixes #616 

@tpolecat Do we want to create a class under the `skunk.exception` or are we happy to use the Exception itself?

### DIFF
--- a/modules/core/shared/src/main/scala/net/BitVectorSocket.scala
+++ b/modules/core/shared/src/main/scala/net/BitVectorSocket.scala
@@ -11,7 +11,7 @@ import fs2.Chunk
 import scodec.bits.BitVector
 import fs2.io.net.{Socket, SocketGroup, SocketOption}
 import com.comcast.ip4s._
-import skunk.exception.EofException
+import skunk.exception.{EofException, SkunkException}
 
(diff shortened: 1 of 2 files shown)


masonedmison:
Couldn't this still fail with an ambiguous error if there is a valid hostname (syntactically) but the host doesn't actually exist ? If I understand correctly, a read on this`Socket` in `sg.client` will still return a `None.get` yielding the `NoSuchElementException`. E.g. `localhosts` would return the previously mentioned exception.

tpolecat (changes requested):
Thanks! I would prefer to raise a `SkunkException` with a detail explaining that the hostname is not syntactically valid and/or that the port falls out of the allowed range. We can fail fast, we don't need a special case for both being invalid.

tpolecat (approved):
Looks great, thanks!

PR #650, opened 2022-04-17
Seperate Initialization of State

ChristopherDavenport:
- Pro: This makes working with this super nice as it doesn't force Kleisli to the edge of your application. This can work inside any lexical scope
- Con: It forces it to be a function since scala 2 doesn't have the ability to return functions requiring implicits.


modules/core/shared/src/main/scala/util/Pool.scala
```diff
@@ -49,11 +49,11 @@ object Pool {
    * @param recycler a cleanup/health-check to be done before elements are returned to the pool;
    *   yielding false here means the element should be freed and removed from the pool.
    */
-  def of[F[_]: Concurrent: Trace, A](
-    rsrc:  Resource[F, A],
+  def of[F[_]: Concurrent, A](
+    rsrc:  Trace[F] => Resource[F, A],
```
mpilquist, line 53: This is binary incompatible, right?

ChristopherDavenport: This should be fixed now with a name change for this, and introducing a universal scoped version thats equivalent to the previous of.

PR #664, opened 2022-06-10
Add scala-xml module

bplommer:


build.sbt
```diff
@@ -198,10 +198,23 @@ lazy val circe = crossProject(JVMPlatform, JSPlatform)
     )
   )
 
+lazy val `scala-xml` = crossProject(JVMPlatform, JSPlatform)
```
armanbilge, line 201: Probably easiest to make it JVM only for now.

bplommer: That requires rearranging the tests too, since they're cross-built for js and jvm. I've tried moving the test into jvm-only sources, let's see what happens.

armanbilge:
```scala
tlVersionIntroduced := List("2.12", "2.13", "3").map(_ -> "skunk version this will be added").toMap
```

build.sbt
armanbilge: Right, nothing wrong with that. Still, better not to publish for JS unless you have a cross-platform parser. scala-xml.js is kind of a lie at the moment.

bplommer: I guess we'd then need to also exclude the js version from mima? Might be better to reorganize the modules instead to do this properly

bplommer:
Note on testing - I've copied the pattern of the Circe tests but they aren't really adequate (they'll still pass if you change the Postgres types of the codecs.) I've done some manual end-to-end testing but we don't have any automated tests that these codecs actually work correctly with Postgres.

modules/scalaxml/src/main/scala/XmlCodecs.scala
```diff
@@ -0,0 +1,20 @@
...
+package scalaxml.codec
+
+import cats.syntax.all._
+
+import scala.xml._
+import skunk.data.{Arr, Type}
+
+trait XmlCodecs {
+  val xml: Codec[Elem] = Codec.simple(_.toString, s => Either.catchNonFatal(XML.loadString(s)).leftMap(_.getMessage), Type.xml)
+  val _xml: Codec[Arr[Elem]] = Codec.array(_.toString, s => Either.catchNonFatal(XML.loadString(s)).leftMap(_.getMessage), Type._xml)
```
armanbilge, line 15: If we use fs2-data-xml for parsing then it can cross-compile for JS/Native.

PR #683, opened 2022-07-18
add read timeout

svalaskevicius:
Adds read timeout for socket reads.

A reason to add it within the library rather than timing out skunk calls - it also applies for the healthcheck queries.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,18 +2,20 @@ version: '3'
 services:
   # main instance for testing
   postgres:
-    image: postgres:11
+    # image: postgres:11
+    build:
+      context: world
     command: -c ssl=on -c ssl_cert_file=/var/lib/postgresql/server.crt -c ssl_key_file=/var/lib/postgresql/server.key
-    volumes:
-      - ./world/world.sql:/docker-entrypoint-initdb.d/world.sql
-      - ./world/server.crt:/var/lib/postgresql/server.crt
-      - ./world/server.key:/var/lib/postgresql/server.key
+    # volumes:
+    #   - ./world/world.sql:/docker-entrypoint-initdb.d/world.sql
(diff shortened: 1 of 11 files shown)


svalaskevicius:
@mpilquist yeah, I noticed that commit, and found no viable alternative means to recover the removed functionality - thus the PR :)

I've rebased on main now, and added a test - please can you have a look?

docker-compose.yml
```diff
@@ -2,18 +2,20 @@ version: '3'
...
+      context: world
     command: -c ssl=on -c ssl_cert_file=/var/lib/postgresql/server.crt -c ssl_key_file=/var/lib/postgresql/server.key
-    volumes:
-      - ./world/world.sql:/docker-entrypoint-initdb.d/world.sql
-      - ./world/server.crt:/var/lib/postgresql/server.crt
-      - ./world/server.key:/var/lib/postgresql/server.key
+    # volumes:
+    #   - ./world/world.sql:/docker-entrypoint-initdb.d/world.sql
+    #   - ./world/server.crt:/var/lib/postgresql/server.crt:ro
+    #   - ./world/server.key:/var/lib/postgresql/server.key:ro
```
svalaskevicius, line 12: I wasn't able to run like this due to mounted key file ownership / permissions within the container - how is it run by other people?

armanbilge: CI does this.
https://github.com/tpolecat/skunk/blob/f67aceec5119e7bf7a9fc3151bfb310bbfee3e54/.github/workflows/ci.yml#L69-L75

svalaskevicius: yeah I did this in the Dockerfile, so I suppose this change should be acceptable? ie.. Dockerfile in `world/` doesn't do more than that :)

PR #684, opened 2022-07-19
Query dynamic rows

### LINKED ISSUE #662: Support `select * from ...` where schema is not known at compile time
(originally asked on discord: https://discord.com/channels/632277896739946517/839844929877704785/977322704363479050)

I need to query a table where the schema is not known at compile time. That means I'd like to do a `select * from dynamicTable` and get 
an `Array[String | null]` for every row, which I can further process according to the types I know at runtime.

Low-level types in the array would also be ok, since I can cast them accordingly. But I need to know if a field is null or not.

So far in ScalaJS, it seems like I'm stuck with with `pg-node` and use their `rowmode: 'array'` feature (https://node-postgres.com/features/queries)

fdietze:
I tried to implement a `.queryDynamic` on `Fragment` with the idea to give me a `Query[A, List[Option[String]]]` for queries, where the schema is not known at compile time. But at runtime, I get an error that the returned rows from postgres don't match the expected types. Any idea?

```
...
a     int4    ->    ── unmapped column
b     bool    ->    ── unmapped column
c     float8  ->    ── unmapped column
...
```

fixes #662

### DIFF
--- a/modules/core/shared/src/main/scala/Fragment.scala
+++ b/modules/core/shared/src/main/scala/Fragment.scala
@@ -28,13 +28,13 @@ final case class Fragment[A](
     } .runA(1).value.combineAll
 
   def query[B](decoder: Decoder[B]): Query[A, B] =
-    Query(sql, origin, encoder, decoder)
+    Query(sql, origin, encoder, decoder, isDynamic = false)
 
   def queryDynamic: Query[A, List[Option[String]]] =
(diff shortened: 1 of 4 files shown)


fdietze:
@armanbilge could you please review? I have this running in production for some time now.

armanbilge:
Thanks for the ping, glad it's working for you :) I'm not really a maintainer here.

armanbilge:
See
- https://github.com/tpolecat/skunk/issues/697

PR #703, opened 2022-09-12
Scala Native

armanbilge:
So this is fun! All tests are passing :) but a couple notes:

1. The build is still a bit hacked, relying on some bootleg deps (enumeratum, natchez, etc.).
2. I made some minor breaking changes and bumped the base version. With some platforming ugliness this could be avoided.


tpolecat:
holy crap

build.sbt
```diff
@@ -142,12 +142,16 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
...
-      "com.armanbilge"  %%% "enumeratum"   % "1.7.1-SNAPSHOT",
-    ).filterNot(_ => tlIsScala3.value)
+    )
   ).jvmSettings(
     libraryDependencies += "com.ongres.scram" % "client" % "2.1",
-  ).platformsSettings(JSPlatform, NativePlatform)(
+  )
+  .platformsSettings(JVMPlatform, JSPlatform)(
+    libraryDependencies ++= Seq(
+      "com.beachape"  %%% "enumeratum"   % "1.7.1-SNAPSHOT",
```
mpilquist: Does this still need to be a snapshot now that's it's moved to JVM/JS

mpilquist: Nevermind, I see that was changed in a subsequent commit.

PR #656, opened 2022-05-05
Extend support for Twiddles to length 16 on Scala 3

benhutchison:
context: https://discord.com/channels/632277896739946517/839844929877704785/970836364746694707

PR #704, opened 2022-09-21
Added Twiddler product instances up to arity 26

vbergeron:
There should be less issue with row decoding.

I view that as a band-aid as a more generic solution base on `NonEmptyTuple` would have been nicer. On my radar but right now the compiler is not kind with me.


tpolecat:
FYI in Scala 3 you can use `.pmap` and `.pcontramap`, which don't need `Twiddler`.

vbergeron:
Thanks for your quick reply. That's perfect :)
Do we want to close this or keep it as a superset of  #656 ?

mpilquist:
@vbergeron Thanks! I lean towards closing this one and sticking with 16 as max built-in arity unless folks really need it for cross-compilation with Scala 2 code using Skunk. For Scala 3 only code bases, we should encourage use of `*:` and `pmap`.

PR #708, opened 2022-09-22
Cross-platform `SSL.Trusted`

armanbilge:
fs2-io.js now supports insecure TLS context, so we can unplatform this code.

Ha, looks like I just missed the release train 😂 eh.


mpilquist:
If it makes you feel better, I'd have preferred this to go in a 0.4 release anyway :)

armanbilge:
Crap, bincompat. If a bump is likely this can just wait for that.

mpilquist:
Yeah, let's bump to 0.4 right after 0.3.2 is published

PR #710, opened 2022-09-22
Use `Resource` for `TLSContext`

armanbilge:
This pulls out the breaking change from the Native PR https://github.com/tpolecat/skunk/pull/703, which may be blocked for a while still on upstream dependencies.

Although this is a fairly breaking change I'm not convinced this is really a user-facing API. Seems like it would be strange to be building `TLSContext`s with Skunk `SSL`.

If we're 👎 on this change, then the Native PR will have to introduce some platforming to work around this, not the end of the world.

### DIFF
--- a/modules/core/shared/src/main/scala/SSL.scala
+++ b/modules/core/shared/src/main/scala/SSL.scala
@@ -5,6 +5,7 @@
 package skunk
 
 import cats._
+import cats.effect.Resource
 import cats.syntax.all._
 import fs2.io.net.Network
 import fs2.io.net.tls.TLSContext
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/SSL.scala
```diff
@@ -16,25 +17,25 @@ abstract class SSL private[skunk] (
   val fallbackOk:    Boolean       = false,
 ) { outer =>
 
-  def tlsContext[F[_]: Network](implicit ev: ApplicativeError[F, Throwable]): F[TLSContext[F]]
+  def tlsContext[F[_]: Network](implicit ev: ApplicativeError[F, Throwable]): Resource[F, TLSContext[F]]
```
armanbilge, line 20: This is the signature change, everything else is catching up with this.

modules/core/shared/src/main/scala/SSL.scala
```diff
@@ -16,25 +17,25 @@ abstract class SSL private[skunk] (
...
 
   def toSSLNegotiationOptions[F[_]: Network](logger: Option[String => F[Unit]])(
     implicit ev: ApplicativeError[F, Throwable]
-  ): F[Option[SSLNegotiation.Options[F]]] =
+  ): Resource[F, Option[SSLNegotiation.Options[F]]] =
     this match {
-      case SSL.None => none.pure[F]
+      case SSL.None => Resource.pure(None)
       case _ => tlsContext.map(SSLNegotiation.Options(_, tlsParameters, fallbackOk, logger).some)
     }
```
armanbilge, line 40: ... which causes this API to change ...

modules/core/shared/src/main/scala/Session.scala
```diff
@@ -281,7 +281,7 @@ object Session {
 
     for {
       dc      <- Resource.eval(Describe.Cache.empty[F](commandCache, queryCache))
-      sslOp   <- Resource.eval(ssl.toSSLNegotiationOptions(if (debug) logger.some else none))
+      sslOp   <- ssl.toSSLNegotiationOptions(if (debug) logger.some else none)
       pool    <- Pool.of(session(Network[F], sslOp, dc), max)(Recyclers.full)
     } yield pool
```
armanbilge, line 286: ... which is ultimately used only here.

PR #712, opened 2022-09-24
Configure codecov to only fail if there is at least 1 percent decrease in coverage

mpilquist:
With the default codecov.io configuration, lots of good builds are marked failed due to slight variance in branch coverage, likely caused by different generated values.


armanbilge:
Codecov is down atm.

https://status.codecov.com/

PR #734, opened 2022-11-03
Fix mdoc markdown processing

sv3ndk:
# Why 

At the moment, the creation of Skunk's documentation with sbt `docs/makeSite` command fails with a `Connection refused` error, as visible in https://github.com/tpolecat/skunk/issues/713#issuecomment-1258124108

The error occurs during the execution of `docs/mdoc` which tries to open a socket when running snippets like `CommandExample.run(Nil).unsafeRunSyncWithRedirect()` 

# What

This PR fixes the issue by marking those snippet as `compile-only`, telling mdoc not to try to execute them. 

# How to test

Running `docs/makeSite` on this branch no longer triggers the connection error and produces a static web site  in `modules/docs/target/paradox/site/main/`

### DIFF
--- a/modules/docs/src/main/paradox/tutorial/Command.md
+++ b/modules/docs/src/main/paradox/tutorial/Command.md
@@ -280,7 +280,7 @@ object CommandExample extends IOApp {
 
 Running this program yields the following.
 
-```scala mdoc:passthrough
+```scala mdoc:compile-only
 println("```")
 import skunk.mdoc._
 CommandExample.run(Nil).unsafeRunSyncWithRedirect()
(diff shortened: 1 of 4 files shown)


armanbilge:
Do you experience this issue if you setup the test db instances? I imagine that's why this works in CI.

https://github.com/tpolecat/skunk/blob/d80f09001fd94cddc60a5bdbea0d2ce111f2d9b8/.github/workflows/ci.yml#L69-L75

sv3ndk:
Hi, thanks for the quick feed-back. 
Oh indeed, starting postgres before generating the docs fixes my issue (I should have thought of that...), I'll close my PR. 
I was trying to help address the fact that the [microsite](https://tpolecat.github.io/skunk/index.html) is not based on the latest release.

PR #761, opened 2022-12-16
Reduce unnecessary string allocation when debug is disabled in MessageSocket

mpilquist:


modules/core/shared/src/main/scala/net/MessageSocket.scala
```diff
@@ -41,11 +41,14 @@ object MessageSocket {
...
   def fromBitVectorSocket[F[_]: Concurrent: Console](
     bvs: BitVectorSocket[F],
-    debug: Boolean
+    debugEnabled: Boolean
   ): F[MessageSocket[F]] =
     Queue.circularBuffer[F, Either[Any, Any]](10).map { cb =>
       new AbstractMessageSocket[F] with MessageSocket[F] {
 
+        private def debug(msg: => String): F[Unit] =
+          Console[F].println(msg).whenA(debugEnabled)
```
ybasket, line 50: Wouldn't that evaluate `msg` still as `println` takes a strict argument? I expected something like
```suggestion
          Applicative[F].whenA(debugEnabled)(Console[F].println(msg))
```
which takes a `=> F[A]`.

armanbilge: I think there is still a thunk allocation here, which we can avoid. There's also the thunk for the string, which we are stuck with unless we reach for macros.
```suggestion
          if (debugEnabled) Console[F].println(msg) else Concurrent[F].unit
```

PR #770, opened 2022-12-26
Small touch: lower evalMap to foreach

diesalbla:
Slightly lighter.

Potentially, we could pass down the queue to `next` to even reduce the type of the stream...


mpilquist:
OBE

diesalbla:
@mpilquist OBE?

PR #791, opened 2023-02-01
WIP: Add beginner session and scala3 codec doc

Quafadas:
I wrote out the documentation, which I believe _may_ have helped me to get Skunk in the right place in my own programs. 

This may be undesirable for several reasons. 
- It may not be best practice - I'm not a confident user of skunk, 
- It may be both too specific (an http4s app) 
- As well as too broad - should skunk docs concern themselves with helping with app design? 

What I'm saying is that I would not be offended, if this documentation PR was not merged - but at least I tried to contribute my learning back.


GitsMcGee:
This documentation would have helped me when I first got started, so thank you!

PR #823, opened 2023-03-20
Misc fixes to allow local file to work with both Docker for Mac and C…

matthughes:
…olima.

With this change, I'm able to run all the Docker containers in Colima short of the redshift one.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,11 +26,7 @@ ThisBuild / nativeBrewInstallCond := Some("matrix.project == 'rootNative'")
 
 lazy val setupCertAndDocker = Seq(
   WorkflowStep.Run(
-    commands = List("export SERVER_KEY=$(cat world/server.key)", "export SERVER_CERT=$(cat world/server.crt)"),
-    name = Some("Set up cert permissions"),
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -26,11 +26,7 @@ ThisBuild / nativeBrewInstallCond := Some("matrix.project == 'rootNative'")
 
 lazy val setupCertAndDocker = Seq(
   WorkflowStep.Run(
-    commands = List("chmod 600 world/server.key", "sudo chown 999 world/server.key"),
-    name = Some("Set up cert permissions"),
-  ),
-  WorkflowStep.Run(
-    commands = List("docker-compose up -d"),
+    commands = List("export SERVER_KEY=$(cat world/server.key)", "export SERVER_CERT=$(cat world/server.crt)", "docker-compose up -d"),
```
armanbilge, line 29: Instead of `export`, should these be set as proper `env = Map(...)` variables for the `WorkflowStep` ?

armanbilge: Oh, sorry, I see, they are reading from a file 🤔

PR #830, opened 2023-03-27
Update to fs2 3.7.0-RC4

mpilquist:


modules/example/src/main/scala/Minimal2.scala
```diff
@@ -53,7 +53,7 @@ object Minimal2 extends IOApp {
       }
     }
 
-  def runF[F[_]: Async: Trace: Parallel: Console]: F[ExitCode] =
+  def runF[F[_]: Async: Trace: Parallel: Console: Network]: F[ExitCode] =
```
armanbilge, line 56: Instead of `parallel` this probably just needs `import cats.effect.implicits._`

armanbilge: https://github.com/tpolecat/skunk/commit/425335900633ca6769293394cfb917c8a0997359 does all of the relevant tidying :)

mpilquist: Looks good, can you PR that?

PR #834, opened 2023-04-01
Sketch of how to improve read performance

mpilquist:


modules/core/shared/src/main/scala/net/Protocol.scala
```diff
@@ -209,7 +209,7 @@ object Protocol {
     readTimeout:  Duration
   ): Resource[F, Protocol[F]] =
     for {
-      bms <- BufferedMessageSocket[F](host, port, 256, debug, sg, socketOptions, sslOptions, readTimeout) // TODO: should we expose the queue size?
```
mpilquist, line 212: This change didn't appear to make a difference so not sure we should touch it at this point.

modules/core/shared/src/main/scala/net/BitVectorSocket.scala
```diff
@@ -48,20 +48,47 @@ object BitVectorSocket {
...
+      private def doRead =
+        withTimeout(socket.read(Int.MaxValue)).flatMap { 
+          case Some(bytes) => carryRef.update(_ ++ bytes.toByteVector.bits)
+          case None => F.unit
+        }
+
+      override def read(nBytes: Int): F[BitVector] = {
+        val nBits = nBytes * 8
+        // TODO mutex & avoid double access to carryRef
+        carryRef.get.flatMap { carry =>
```
armanbilge, line 82: Maybe `carryRef` should be an `AtomicCell` ?

https://typelevel.org/cats-effect/docs/std/atomic-cell

mpilquist: Ha, nice! Definitely `AtomicCell` if we need synchronization. Not sure we do in this case though -- adding it makes test about 33% slower. Thinking maybe we should drop `BitVectorSocket` entirely or at least make `MessageSocket` smarter about performing reads -- e.g., build a whole `Decoder[Message]` instead of reading a few bytes, decoding a few bytes, etc.

PR #835, opened 2023-04-02
Improve performance of reading large result sets

### LINKED ISSUE #833: Retrieval performance
Skunk seems to be orders of magnitude slower than psql when executing simple read operations on large datasets.  

Using `select generate_series(1,500000)` as our query, I tested using psql and Skunk.  Running tests
- M1 Mac
- using Postgres 11 docker image from Skunk tests
- running Skunk test with 8gb allocated to sbt

## psql test

- open psql 
- dump output to a file '\o scratch.txt'
- turn on timings via `\timing`
- run query
- takes 115ms

## skunk test

I added some tests to QueryTest:

```scala
    // Takes ~12s 
    sessionTest("large row execute benchmark") { s => 
      val query = sql"""select generate_series(1,500000)""".query(int4)
      for {
        res <- s.execute(query).timed
        (duration, r) = res
        _ = println(s"Took ${duration.toMillis} to return 500K rows")
        _ <- assertEqual(500_000.toString, 500_000, r.size)
      } yield "ok"
    }

    // Even when streaming the results and not accumulating them, we see same time ~12s 
    sessionTest("large row stream benchmark") { s => 
      val query = sql"""select generate_series(1,500000)""".query(int4)
      for {
        res <- s.stream(query, Void, 64_000).compile.drain.timed
        (duration, r) = res
        _ = println(s"Took ${duration.toMillis} to stream 500K rows to /dev/null")
      } yield "ok"
    }
```

Even when not accumulating all 500K rows in memory (second test) we still see results 100x slower than via psql.

mpilquist:
Fixes #833

### DIFF
--- a/modules/core/shared/src/main/scala/net/BitVectorSocket.scala
+++ b/modules/core/shared/src/main/scala/net/BitVectorSocket.scala
@@ -35,31 +35,30 @@ object BitVectorSocket {
    * @param socket the underlying `Socket`
    * @group Constructors
    */
-  def fromSocket[F[_]](
-    socket:       Socket[F],
-    readTimeout:  Duration
-  )(
-    implicit ev: Temporal[F]
-  ): BitVectorSocket[F] =
+  def fromSocket[F[_]](socket: Socket[F], readTimeout: Duration, carryRef: Ref[F, Chunk[Byte]])(implicit F: Temporal[F]): BitVectorSocket[F] =
     new BitVectorSocket[F] {
-
-      val withTimeout: F[Chunk[Byte]] => F[Chunk[Byte]] = readTimeout match {
+      private val withTimeout: F[Option[Chunk[Byte]]] => F[Option[Chunk[Byte]]] = readTimeout match {
         case _: Duration.Infinite   => identity
         case finite: FiniteDuration => _.timeout(finite)
       }
 
-      def readBytes(n: Int): F[Array[Byte]] =
-        withTimeout(socket.readN(n)).flatMap { c =>
-          if (c.size == n) c.toArray.pure[F]
-          else ev.raiseError(EofException(n, c.size))
+      override def read(nBytes: Int): F[BitVector] =
+        // nb: unsafe for concurrent reads but protected by protocol mutex
+        carryRef.get.flatMap(carry => readUntilN(nBytes, carry))
+
+      private def readUntilN(nBytes: Int, carry: Chunk[Byte]): F[BitVector] =
+        if (carry.size < nBytes) {
+          withTimeout(socket.read(8192)).flatMap {
+            case Some(bytes) => readUntilN(nBytes, carry ++ bytes)
+            case None => F.raiseError(EofException(nBytes, carry.size.toInt))
+          }
+        } else {
+          val (output, remainder) = carry.splitAt(nBytes)
+          carryRef.set(remainder).as(output.toBitVector)
         }
 
-      override def read(nBytes: Int): F[BitVector] =
-        readBytes(nBytes).map(BitVector(_))
-
       override def write(bits: BitVector): F[Unit] =
-        socket.write(Chunk.array(bits.toByteArray))
-
+        socket.write(Chunk.byteVector(bits.bytes))
     }
 
   /**
@@ -91,7 +90,8 @@ object BitVectorSocket {
     for {
       sock <- sock
       sockʹ <- sslOptions.fold(sock.pure[Resource[F, *]])(SSLNegotiation.negotiateSSL(sock, _))
-    } yield fromSocket(sockʹ, readTimeout)
+      carry <- Resource.eval(Ref.of(Chunk.empty[Byte]))
+    } yield fromSocket(sockʹ, readTimeout, carry)
 
   }
 }
--- a/modules/core/shared/src/main/scala/net/MessageSocket.scala
+++ b/modules/core/shared/src/main/scala/net/MessageSocket.scala
@@ -10,8 +10,6 @@ import cats.effect.std.Console
 import cats.effect.std.Queue
 import cats.syntax.all._
 import scala.io.AnsiColor
-import scodec.codecs._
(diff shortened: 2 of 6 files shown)


modules/core/shared/src/main/scala/net/BitVectorSocket.scala
```diff
@@ -35,31 +35,30 @@ object BitVectorSocket {
...
-    implicit ev: Temporal[F]
-  ): BitVectorSocket[F] =
+  def fromSocket[F[_]](socket: Socket[F], readTimeout: Duration, carryRef: Ref[F, Chunk[Byte]])(implicit F: Temporal[F]): BitVectorSocket[F] =
     new BitVectorSocket[F] {
-
-      val withTimeout: F[Chunk[Byte]] => F[Chunk[Byte]] = readTimeout match {
+      private val withTimeout: F[Option[Chunk[Byte]]] => F[Option[Chunk[Byte]]] = readTimeout match {
         case _: Duration.Infinite   => identity
         case finite: FiniteDuration => _.timeout(finite)
       }
```
armanbilge, line 43: `.timeout(...)` now supports `Duration.Infinite` so this helper shouldn't be needed anymore.

mpilquist: Yeah I tried that but it benchmarked a bit slower since the short circuit happens once with current implementation but on each call with the new CE support.

armanbilge: I'm amazed that makes such a difference 😅

modules/core/shared/src/main/scala/net/MessageSocket.scala
```diff
@@ -54,9 +52,9 @@ object MessageSocket {
         * total including self but not including the tag) in network order.
         */
         val receiveImpl: F[BackendMessage] = {
-          val header = (byte.asDecoder, int32.asDecoder).tupled
           bvs.read(5).flatMap { bits =>
-            val (tag, len) = header.decodeValue(bits).require
+            val tag = bits.take(8).toByte()
```
matthughes, line 56: Are you just inlining the header decoder?  Or why this change?

modules/tests/jvm/src/test/scala/skunk/LargeResponseTest.scala
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2018-2021 by Rob Norris
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package test
```
matthughes, line 5: I think this should be `tests` right?  A

modules/core/shared/src/main/scala/net/BitVectorSocket.scala
```diff
@@ -35,31 +35,30 @@ object BitVectorSocket {
...
-        }
-
       override def read(nBytes: Int): F[BitVector] =
-        readBytes(nBytes).map(BitVector(_))
+        // nb: unsafe for concurrent reads but protected by protocol mutex
+        carryRef.get.flatMap(carry => readUntilN(nBytes, carry))
+
+      private def readUntilN(nBytes: Int, carry: Chunk[Byte]): F[BitVector] =
+        if (carry.size < nBytes) {
+          withTimeout(socket.read(8192)).flatMap {
```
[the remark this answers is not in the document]
matthughes: I played with increasing it and it seemed to make no difference fwiw.

modules/core/shared/src/main/scala/net/MessageSocket.scala
mpilquist: Avoids allocations of a bunch of intermediate objects. Makes a small bit consistent improvement on the benchmark.

modules/core/shared/src/main/scala/net/BitVectorSocket.scala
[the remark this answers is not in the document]
mpilquist: Did that too :)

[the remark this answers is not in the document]
guidoschmidt17: I hope it's ok that I add my comments just a visitor.
I noticed a performance improvement only when setting SND/RCV buffers to a really huge value (8mb). This assumes that the session pool is reasonably small (8-64, for instance.). And it's probably a good idea to have it configurable.

modules/tests/jvm/src/test/scala/skunk/LargeResponseTest.scala
guidoschmidt17: Maybe this is too daring: BitVector is used to decode messages in the code above (and of course elsewhere). Maybe it would improve performance using ByteV
ector. To my knowledge the PG wire protocol's smallest unit is one byte. And ByteVector is really highly optimized. It would require a huge refactoring, though. Just my 2ct.

mpilquist: There shouldn’t be a significant difference and scodec’s Codec abstraction is built on BitVector.

modules/core/shared/src/main/scala/net/BitVectorSocket.scala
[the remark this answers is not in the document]
mpilquist: Comments very welcome! The socket options are configurable now. The read size here isn’t but we could make it so. I’d like to show a case where it matters though before breaking compatibility by introducing a new param. Same for the background queue size in BufferedMessageSocket, which is hardcoded to 256 and doesn’t seem to fill.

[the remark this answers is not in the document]
mpilquist: Also, I suspect receive buffer size may be more influential on performance when connecting to a remote postgres database as opposed to local host.

matthughes (approved):

PR #836, opened 2023-04-03
Better composition for fragments

vbergeron:
This PR aim to improve developer UX when composing fragments.

The sql interpolator is modified to remove optimize more the composition of `Void` codecs.
* `Void ~ Void -> Void`
* `A    ~ Void -> A`
* `Void ~ B    -> B`
* `A    ~ B    -> A ~ B`


mpilquist:
@vbergeron Mind fixing the merge conflicts on this one? Also, we should think about how to handle source compatibility here. Now that we have `skunk.feature.legacyCommandSyntax`, perhaps we should leave `Void` (and `Unit`?) values in place when legacy syntax is enabled.

PR #845, opened 2023-04-17
Make query decoder implicit

DeamonDev:
The aim of this PR is to make `decoder` inside `Fragment.query` method an implicit parameter.

### DIFF
--- a/modules/core/shared/src/main/scala/Fragment.scala
+++ b/modules/core/shared/src/main/scala/Fragment.scala
@@ -27,7 +27,7 @@ final case class Fragment[A](
       case Right(s) => s
     } .runA(1).value.combineAll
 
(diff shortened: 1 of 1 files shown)


mpilquist:
Thanks for the PR! I don't think we should merge this one though. Treating codecs as plain values instead of implicit values is a deliberate design choice.

Some background here: https://youtu.be/NJrgj1vQeAI?t=3018

We might add support for derivation in the future but if so, we'd do so in a way that keeps the recommended approach explicit codecs.

DeamonDev:
Thanks for response! I wasn't aware of that discussion.

mpilquist:
No worries at all! This comes up often - we should document it in an FAQ or somewhere else in docs.

PR #846, opened 2023-04-17
Use Typelevel Twiddles

mpilquist:
This PR deprecates left-nested twiddle lists in favor of using [Typelevel Twiddles](https://github.com/typeleve/twiddles).

Typelevel Twiddles provides a single API that:
- works on both Scala 3 and Scala 2, allowing cross compilation of client code
- supports arbitrary arity (i.e. not limited to 22 elements)
- provides seamless support for regular tuples on Scala 3 and compatibility with regular tuples on Scala 2

As a result, codecs/encoders/decoders are now composed via `*:` instead of `~` and `gimap`/`gmap`/`gcontramap` have been replaced with `as`.

The `~` operation is deprecated and will be removed in Skunk 1.0.

Feats of strength were performed to maintain source compatibility. Source breakage was unavoidable in the case of parameterized commands, which now return commands of new style twiddle lists. See the docs page on Twiddle Lists for migration, including the option of adding `import skunk.feature.legacyCommandSyntax` to (temporarily) restore the old behavior.


mpilquist:
Any thoughts on when (if?) we should deprecate legacy twiddle lists (i.e. the left nested tuples). My initial thought was to deprecate them in the 0.6 release and remove them entirely in 1.0. However, parameterized commands make that gradual cutover difficult. Consider:

```scala
  val insertCity2: Command[City] =
    sql"""
        INSERT INTO city
        VALUES ($int4, $varchar, ${bpchar(3)}, $varchar, $int4)
      """.command.contramap {
            case c => c.id ~ c.name ~ c.code ~ c.district ~ c.pop
          }
```

The `sql` interpolator uses `~` for params. If we change that to use `*:` instead, we break all usage like the above.

chuwy:
I humbly vote for deprecating them ASAP. The `*:` / `~` ambiguity caused lots of problems in my derivation code (I have such thing in my private lib).

mpilquist:
One option that I haven't fully thought through is adding an import that opts-in to the 1.0 syntax.

For example, without an explicit import, this could would continue to work without change:

```scala
  val insertCity2: Command[City] =
    sql"""
        INSERT INTO city
        VALUES ($int4, $varchar, ${bpchar(3)}, $varchar, $int4)
      """.command.contramap {
            case c => c.id ~ c.name ~ c.code ~ c.district ~ c.pop
          }
```

But if you added `import skunk.syntax.future`, then you'd get to write:

```scala
  // On Scala 3 only
  val insertCity2: Command[City] =
    sql"""
        INSERT INTO city
        VALUES ($int4, $varchar, ${bpchar(3)}, $varchar, $int4)
      """.command.contramap {
            case c => (c.id, c.name, c.code, c.district, c.pop)
          }

  // On Scala 2 or Scala2+3 cross build
  val insertCity2: Command[City] =
    sql"""
        INSERT INTO city
        VALUES ($int4, $varchar, ${bpchar(3)}, $varchar, $int4)
      """.command.contramap {
            case c => c.id *: c.name *: c.code *: c.district *: c.pop *: EmptyTuple
          }
````

**Comment by michael-smith** [comment]
> One option that I haven't fully thought through is adding an import that opts-in to the 1.0 syntax.
> 

Never really been a fan of the future opt-in import.  I agree with @chuwy that it should just be dealt with ASAP.

File: modules/core/shared/src/main/scala-2/SkunkPlatform.scala [hunk 1]
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2018-2021 by Rob Norris
+// This software is licensed under the MIT License (MIT).
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package skunk
+
+private[skunk] trait SkunkPlatform {
```
armanbilge, line 7: Sorry, nit, trying to reserve `Platform` for platform concerns. Hopefully would never be needed...

```suggestion
private[skunk] trait SkunkCompat {
```

mpilquist:
Commands now default to the new twiddle syntax. Adding `import skunk.feature.legacyCommandSyntax` results in commands using left nested tuples.

PR #853, opened 2023-04-22
Use `Mutex` instead of `Semaphore(1)`

armanbilge:


modules/core/shared/src/main/scala/net/protocol/Exchange.scala
```diff
@@ -15,10 +15,10 @@ trait Exchange[F[_]] {
 
 object Exchange {
   def apply[F[_]: Concurrent]: F[Exchange[F]] =
-    Semaphore[F](1).map { sem =>
+    Mutex[F].map { mut =>
       new Exchange[F] {
         override def apply[A](fa: F[A]): F[A] =
-        sem.permit.use(_ => fa).uncancelable
+          mut.lock.surround(fa).uncancelable
```
armanbilge, line 21: While we're here, I wonder about this being `.uncancelable`. Should the lock acquisition be uncancelable as well? 🤔 or would `mut.lock.surround(fa.uncancelable)` be sufficient.

PR #855, opened 2023-04-26
completion for `set constraints`

tpolecat:
I can't figure out how to run tests locally anymore so let's see if this works.


mpilquist:
@tpolecat Are you on OS X? x86 or ARM? Things should mostly just work if using Docker Desktop & Intel. The Redshift tests don't work on ARM.

PR #857, opened 2023-04-27
otel4s-0.2.1

rossabaker:
Continuation of https://github.com/rossabaker/skunk/pull/2.  Replaces Natchez tracing with otel4s.


rossabaker:
Docs don't compile yet.  There are a few compilation warnings that I don't think I introduced.  Just wanted to get it out of the shadows.

rossabaker:
docker-compose is trying to make directories in my root folder, so I can't easily run `mdoc`.  There's a lot of tedious `noop` work to be done, and I think otel4s should adopt Natchez's approach to that.

otel4s isn't published for 2.12.  I haven't kept up with Skunk's plans on that front.

mpilquist:
Re: 2.12, let's drop it for Skunk 1.0.

mpilquist:
I pushed a few commits that should get build working (or at least closer).

rossabaker:
Sounds good to me.  There are a couple improvements I'd like to make to otel4s from this experience, but better to get the big one merged and refine later.

PR #872, opened 2023-05-21
Make the example compiles

lenguyenthanh:


modules/docs/src/main/laika/tutorial/Setup.md
```diff
@@ -36,11 +36,10 @@ import skunk._
 import skunk.implicits._
 import skunk.codec.all._
 import org.typelevel.otel4s.trace.Tracer
+import natchez.Trace.Implicits.noop                          // (1)
```
armanbilge, line 39: Hmm, we shouldn't be replacing otel4s with natchez in the main branch. I think the problem is that the doc site is currently published from main instead of the series/0.6 branch.

lenguyenthanh: I think it is published from `0.6` branch because I saw `0.6.0-RC2` in the website. btw is the anyway for me to use skunk without otel4s? I saw log4cats but haven't tried it yet. Another thing is: the version in otel4s also doesn't work: `"org.typelevel" %% "otel4s-java" % "0.3-581687e-SNAPSHOT"` I couldn't find this version, and have to use `0.2.1` instead.

armanbilge: > I think it is published from `0.6` branch because I saw `0.6.0-RC2` in the website.

no, that's just because there are no published tags on the main branch.

> btw is the anyway for me to use skunk without otel4s?

yes, all currently published versions of skunk are based on natchez. currently there is no published version which uses otel4s.

armanbilge: btw, I opened a PR to fix that.
- https://github.com/typelevel/skunk/pull/873

lenguyenthanh: thanks, just saw that. I guess we can close this pr then.

mpilquist:
Closing in favor of #873

PR #874, opened 2023-05-23
Add cases for sql EXPLAIN command, and a unit test.

lucianojoublanc-da:
I had some reservation about the unit test, would appreciate if somebody would check it's robust enough.

Also, I've actually been using `EXPLAIN (FORMAT JSON) ...` and decoding with `skunk-circe`, but it doesn't make sense to add a dependency just for this into the `tests` module, so I don't actually cover this in the unit test.

### DIFF
--- a/modules/tests/shared/src/test/scala/CommandTest.scala
+++ b/modules/tests/shared/src/test/scala/CommandTest.scala
@@ -94,11 +94,6 @@ class CommandTest extends SkunkTest {
          WHERE id = $int4
        """.command
 
-  val explainCity: Query[Void, String] =
-    sql"""
-          EXPLAIN SELECT * FROM city
-        """.query(skunk.codec.all.text)
-
   val createTable: Command[Void] =
     sql"""
       CREATE TABLE IF NOT EXISTS earth (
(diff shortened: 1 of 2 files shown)


modules/tests/shared/src/test/scala/CommandTest.scala
```diff
@@ -94,6 +94,11 @@ class CommandTest extends SkunkTest {
          WHERE id = $int4
        """.command
 
+  val explainCity: Query[Void, String] =
```
mpilquist, line 97: Any reason to put this in CommandTest.scala instead of QueryTest.scala?

lucianojoublanc-da: Looking at the content of the files, `QueryTest.scala` seems to be testing the methods on `Query` class, whereas `CommandTest.scala` appears to be testing SQL statements in general (maybe it would be more logical to call it `StatementTest.scala`).
It seems more appropriate to keep it there, but happy to move to `QueryTest.scala` if you prefer; let me know :+1:

lucianojoublanc-da: Moved test into `QueryTest`.

modules/tests/shared/src/test/scala/QueryTest.scala
```diff
@@ -2,7 +2,7 @@
 // This software is licensed under the MIT License (MIT).
 // For more information see LICENSE or https://opensource.org/licenses/MIT
 
-package test
+package tests
```
lucianojoublanc-da, line 5: I'm assuming this is a typo as all the other packages here are under 'tests' (though the filesystem path is `test`).

PR #878, opened 2023-05-25
Use `skunk-core` artifact in Setup docs

armanbilge:
Not sure why it specifically had `skunk-refined`, which is compounded by the fact this artifact wasn't publishing properly until the next release.


armanbilge:
Oh dang, meant to target this at 0.6.

PR #897, opened 2023-06-13
Update local setup steps of contributing guide

cosminci:


mpilquist:
Hm, did the `bin/local` script not work for you?

cosminci:
I had tried a couple of times and got `Connection refused` when running the tests even though the containers were up (checked even with `docker ps`). 

Tried again now and it works 🙈

PR #898, opened 2023-06-13
Extended query tests

cosminci:
Test `parameterized w/ list` fails. 
It's equivalent to the second test which uses the legacy twiddle syntax.
The first test passes so the issue seems to appear when list (maybe other?) codec is involved.


milessabin:
I don't think this is a compiler bug. The desugaring for right associated calls involves an assignment to an intermediate val (`rassoc$6` in this case) and this is enough to break the identity you're expecting. You can reproduce something similar in both Scala 2 and 3,

```scala
miles@tarski:~% scala-cli --java-home /usr/lib/jvm/java-20-jdk
Welcome to Scala 3.3.0 (20, Java Java HotSpot(TM) 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> val foo = new Object
val foo: Object = java.lang.Object@5cc1bf20

scala> val bar = foo
val bar: Object = java.lang.Object@5cc1bf20

scala> val baz: bar.type = foo
-- [E007] Type Mismatch Error: -------------------------------------------------
1 |val baz: bar.type = foo
  |                    ^^^
  |                    Found:    (foo : Object)
  |                    Required: (bar : Object)
  |
  | longer explanation available when compiling with `-explain`
1 error found

scala> val quux: foo.type = foo
val quux: foo.type = java.lang.Object@5cc1bf20
```

So, unfortunately the _name_ (ie. the precise Symbol) that the value is presented as matters and has to match between the value and the singleton type.

cosminci:
Thanks for looking into this @mpilquist!

I've updated the test to use the explicit twiddle constructor.
Also changed the first test to use twiddles instead of cons and removed the `shapeless._` import.

LMK if we want this merged, or rather add some examples to the docs, or both.

cosminci:
P.S.: If the list codec is not the first, `new *:` constructors need to be chained all the way up starting from it, even if the previous codecs are simple.

cosminci:
@mpilquist do we eventually want to merge this suite for regression purposes, or should I close the PR since it's achieved its purpose?

mpilquist:
Yeah, I think we should include.

cosminci:
My bad, should have ran the tests locally for both 2.13 and 3.

Compilations fails on Scala 3 with:
```
[error] -- Error: /home/runner/work/skunk/skunk/modules/tests/shared/src/test/scala/ExtendedQueryTest.scala:59:45 
[error] 59 |      country <- preparedQuery.stream(new *:(continents, 10_000_000 *: EmptyTuple), chunkSize = 5)
[error]    |                                             ^^^^^^^^^^
[error]    |     too many arguments for constructor *: in class *:: (): Any *: Tuple
```

Trying to find a workaround.

cosminci:
Can't seem to find a way to please both Scala 2.13 and Scala 3 compilers.

`val params: continents.type *: Int *: EmptyTuple = continents *: 10_000_000 *: EmptyTuple`.
^ This syntax is accepted by Scala 3, but not by Scala 2.13.
```
[error]  found   : rassoc$6.type (with underlying type List[String])
[error]  required: continents.type
```

`val params: continents.type *: Int *: EmptyTuple = new *:(continents, new *:(10_000_000 *: EmptyTuple))`
^ This syntax is accepted by Scala 2.13, but not by Scala 3.
```
[error] 58 |    val params: continents.type *: Int *: EmptyTuple = new *:(continents, new *:(10_000_000 *: EmptyTuple))
[error]    |                                                              ^^^^^^^^^^
[error]    |           too many arguments for constructor *: in class *:: (): H *: T
```

I've tried variations, as well as using plain Shapeless, but couldn't find anything that pleases both compilers.

At this point, I feel like I'm just trying random syntax, since I don't have sufficient knowledge of Scala's type system internals.

cosminci:
I couldn't find a way to please both Scala 2.13 and Scala 3 compilers.

```
val params: continents.type :: Int :: HNil =
    new ::[continents.type, Int :: HNil](continents, 10_000_000 :: HNil)
```

^ This isn't accepted by the Scala 3 compiler :(
```
[error]    |too many arguments for constructor *: in class *:: (): ((continents : List[String]), Int)
```

I don't feel like I have sufficient knowledge of the scala type system internals to understand what's happening here.

PR #899, opened 2023-06-13
Flesh out Refined module a bit ( 2 years later :D)

lenguyenthanh:
Just cherry pick commit of @LukaJCB (thanks for your work <3) from #581 . It seems just work. Please tell me if I need to "refine" anything.


lenguyenthanh:
hi @mpilquist, just wonder what do you think about this pr? Do I need to do any thing to push this forward? If not I think we can just use the old pr because I haven't changed anything from the original implementation.

mpilquist:
Sorry for delay on this one!

PR #900, opened 2023-06-13
Add redaction support to encoder and change SkunkException to use it

mpilquist:
Adds support for both encoder level redaction (e.g., `int4.redacted ~ int4`) as well as session level overrides (via `redactionStrategy = RedactionStrategy.All` or `RedactionStrategy.None`).

Some libraries default to redacting all encoded values but in this PR, we default to respecting individual encoder redaction.

This PR also updates the test infrastructure to trace all database tests (using a new top level span per test).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,3 +23,5 @@ To run a specific test:
 > sbt
 > test:testOnly tests.simulation.StartupSimTest
 ```
+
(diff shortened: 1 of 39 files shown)


tpolecat:
I think this is going to work well. Thanks!

mpilquist:
This one is ready for review.

Issue #553, opened 2021-09-23
Concurrent use of portals and simple queries invalidates portals


hnaderi:
To reproduce:

```scala
//from CursorTest.scala 
  sessionTest("invalidation") { s =>
    cursor(s).use { c =>
      for {
        _ <- c.fetch(1)
        _ <- s.execute(sql"select 1;".query(int4))
        _ <- c.fetch(1)
      } yield ()
    }
  }
```

This will fail with exception Problem: Portal "portal_2" does not exist.

From what is seen on tcpdump and wireshark, Portal get invalidated implicitly and server takes it as a given and not sends any message for its invalidation.

expected behavior: 
This is hard to say, as it is exactly what postgres do which is inconvenient IMO, but this is issue can be mitigated by warnings on document, method docs etc.

tpolecat:
By the way you can add `debug = true` to your `Session` configuration and you'll get a transcript of all the messages that go between client and server. No need for Wireshark, etc.

hnaderi:
related document which specifies this behavior:
https://www.postgresql.org/docs/10/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY

> If successfully created, a named portal object lasts till the end of the current transaction, unless explicitly destroyed.

hnaderi:
> No need for Wireshark, etc.
I was trying to capture server communication itself to see if other clients behaviors are different, but it turns out to be fruitless :smile:

benhutchison:
I suspect Im hitting this or a related problem. 

Im trying to do a bunch of bulk inserts.

When I do them serially with `traverse` it works. When I try in parallel with `parTraverseN(8)`, I get postgres errors like `Portal "portal_35" does not exist.`

tpolecat:
I would need to see some code I guess, but it sounds like you need to operate inside a transaction if you want to have multiple active portals.

benhutchison:
Thanks for the pointer, issue solved :)

I added a session pool and ran each of my parallel insert jobs in a separate session & transaction. Stable and _very_ fast! 🏎️

matthughes:
Don't the Postgres docs indicate the prepared statement should be good for the life of the connection?

[...]

PR #904, opened 2023-06-17
Add test case that exhibits portal does not exist problem.

matthughes:
In trying to track down the "portal does not exist" problem #553, I was able to create a test case that demonstrates the issue.

My code had a weird construction of .map(effect).flatMap(identity) instead of flatMap(effect).  I was able to reliably reproduce the bug in my code with the former and was never able to reproduce with the latter.  I was able to replicate this in Skunk code.

- flatMap(effect) no issue
- map(effect).flatMap(identity) inside of resource acquisition, no issue
- map(effect) with flatMap(identity) outside of resource acquisition, portal not found
- map(effect) with flatten outside of resource acquisition, portal not found

### DIFF
--- a/modules/tests/shared/src/test/scala/DescribeCacheTest.scala
+++ b/modules/tests/shared/src/test/scala/DescribeCacheTest.scala
@@ -128,6 +128,21 @@ class DescribeCacheTest extends SkunkTest(true) {
       }.flatMap(identity)
     }
   }
+  
+  // This will fail if run enough 
+  pooledTest("portal4 - concurrent portal with flatten *outside tx*") { pool =>
+    import scala.concurrent.duration._
+    def cmd(idx: Int): Command[String *: Int *: EmptyTuple] = sql"INSERT INTO scalars VALUES (${text.varchar}, ${int4}) -- #${idx.toString}".command
+    1.to(runs).toList.parTraverse{ idx =>
+      pool.use { s =>
+        s.transaction.use { _ =>
+          s.prepare(cmd(idx)).flatMap(_.execute((s"name$idx", idx))).map (_ => 
+            IO.sleep(1.second) >> s.prepare(cmd(idx)).flatMap(_.execute((s"name$idx", idx))).void
+          )
+        }
+      }.flatten
+    }
+  }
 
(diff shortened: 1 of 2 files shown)


modules/tests/shared/src/test/scala/DescribeCacheTest.scala
```diff
@@ -82,6 +83,67 @@ class DescribeCacheTest extends SkunkTest {
...
+    import scala.concurrent.duration._
+    def cmd(idx: Int): Command[String *: Int *: EmptyTuple] = sql"INSERT INTO scalars VALUES (${text.varchar}, ${int4}) -- #${idx.toString}".command
+    1.to(runs).toList.parTraverse{ idx =>
+      pool.use { s =>
+        s.transaction.use { _ =>
+          s.prepare(cmd(idx)).flatMap(_.execute((s"name$idx", idx))).map (_ => 
+            IO.sleep(1.second) >> s.prepare(cmd(idx)).flatMap(_.execute((s"name$idx", idx))).void
+          )
+        }
+      }.flatten
```
armanbilge, line 143: `.use(...).flatten` looks wrong. It means you are building an `IO` in the `use(...)` when the `Resource` is available, but you are executing it _after_ the `use`, when the resource has been released.

modules/tests/shared/src/test/scala/DescribeCacheTest.scala
```diff
@@ -82,6 +83,67 @@ class DescribeCacheTest extends SkunkTest {
...
+    import scala.concurrent.duration._
+    def cmd(idx: Int): Command[String *: Int *: EmptyTuple] = sql"INSERT INTO scalars VALUES (${text.varchar}, ${int4}) -- #${idx.toString}".command
+    1.to(runs).toList.parTraverse{ idx =>
+      pool.use { s =>
+        s.transaction.use { _ =>
+          s.prepare(cmd(idx)).flatMap(_.execute((s"name$idx", idx))).map (_ => 
+            IO.sleep(1.second) >> s.prepare(cmd(idx)).flatMap(_.execute((s"name$idx", idx))).void
+          )
+        }
+      }.flatMap(identity)
```
armanbilge, line 128: Same here.

PR #910, opened 2023-06-25
docs: add required import for Scala 2

buntec:


modules/docs/src/main/laika/reference/TwiddleLists.md
```diff
@@ -3,6 +3,8 @@
 Twiddle lists are tuples that are built incrementally, element by element, via the `*:` operation. Skunk uses the [Typelevel Twiddles](https://github.com/typelevel/twiddles) library for parameter encoders and row decoders, and via this mechanism twiddle list types appear as type arguments for `Query` and `Command`.
 
 ```scala
+import org.typelevel.twiddles._ // required for Scala 2
```
mpilquist, line 6: Can we move this import down to the example that defines `t1`, `t2`, etc?

buntec: Of course!

PR #974, opened 2023-09-11
Use explicit Locale for era formatting

busti:
The field `eraFormatter` in `TemporalCodecs.scala` is missing an explicit reference to `Locale.US` which can cause unexpected behavior when a locale db with i.e. british locale is added to the dependencies.  

I encountered this because we added https://github.com/cquiroz/scala-java-locales to our project, whose default db, which is recommended in the repository, is for british english.


armanbilge:
Thanks, should we target this to `series/0.6.x` branch?

busti:
Specifically this arose because we were using `$date` from `TemporalCodecs` in a query which before adding the library formatted a date as `2023-03-12` and as `2023-03-12 CE` after which postgres wasn't able to process.

Issue #959, opened 2023-08-18
Feature reqest: migrations. 


KristianAN:
Projects like Flyway and Liquibase enable schema creation and migration. They are however not multiplatform. It would be nice it Skunk provided a form of migration tooling. 

I don't think there is any point in implementing all the functionality of the mentioned projects, but just the ability to perform simple forward migrations would be a great QoL feature (in my opinion). It allows for easy deployments and makes it easy to keep the schema in sync when developing in a team.

KristianAN:
There is this library https://github.com/rolang/dumbo by @rolang . Dumbo is however tied to the flyway spec, which some might not like because of flyways open-core model?

Issue #695, opened 2022-08-08
Support for multi-query statements


matthughes:
The postgres protocol supports sending multiple queries across different tables/views (separated by semi-colon) in one request and receiving the results.  Currently Skunk explicitly does not support this:

From Query.scala
```scala
          case ReadyForQuery(_) =>
            new SkunkException(
              message   = "Multi-statement queries and commands are not supported.",
              hint      = Some("Break up semicolon-separated statements and execute them independently."),
              sql       = Some(stmt.sql),
              sqlOrigin = Some(stmt.origin),
            ).raiseError[F, Unit].whenA(multipleStatements)
```

Even if you have parallelizeable queries in a transaction, currently clients are forced to submit the queries one at a time, waiting for the results of the first query before submitting the second query.  If you have a lot of queries and a decent amount of latency between your app/db servers this can add up.  

I'm interested in adding support for this and am digging around but wouldn't mind if anyone had any pointers or reasons why this could never work with the Skunk model.

tpolecat:
How would you imagine it looking? What would the type of such a query be?

rolang:
Hello, I had another use case for this, just to provide some input. I am trying to replace [flyway](https://flywaydb.org/) that we use for forward-migrations only in our codebase using skunk and bootstrapped [dumbo](https://github.com/rolang/dumbo).

I started by reading the files which can contain multiple statements and then just `fileContent.split(';')` to split it into single-query statements which worked fine until I encountered cases with like `CREATE FUNCTION` statements having `;` all over the place which made it more difficult to parse.

Later I found out that one can just send the whole file content to Postgres as it supports multi-query statements which made it much easier. In this case I was interested on running multiple statements within a transaction and don't really care about the output as long as it is not an error, similar to how Flyway would handle it. Flyway also doesn't care whether you run queries or commands in your migration script, as long as there is no error from Postgres it will execute it and update the migration state.

I currently applied a workaround, basically by changing the signature of finishUp to keep accumulating received backend messages instead of failing together with a multiple completion case `case class Multiple(list: List[Completion]) extends Completion`. It did the job for this case, but not sure about how this could be modelled well in skunk.

```scala
// from
def finishUp(stmt: Statement[_], multipleStatements: Boolean = false): F[Unit]
// to
def finishUp(stmt: Statement[?], messages: List[BackendMessage] = Nil): F[List[BackendMessage]]
```
Sources:
[...]

PR #1023, opened 2023-12-30
Support for execution of any single or multi query statements with discarded completions/rows

rolang:
Hi, I was looking for a way to execute any statements without being able to know in advance whether it's going to be a query or command as well as whether it'll contain single or multiple queries which is currently not possible.

Use case with some background:
We are using skunk in our code base for database access which has been working out great. For running migrations / versioning the database though we are still seeking support from [flyway](https://flywaydb.org). Considering that we only use a fraction of Flyway features, only with Postgres and we already have a library for database access which could take care of that, it adds a lot of dependencies on top like JDBC with drivers and keeps us tied to the JVM. Flyway cli also ships with a jre.

I am attempting to replace Flyway with [dumbo](https://github.com/rolang/dumbo). For that I'd like to read any SQL statements from a file, send it to Postgres and let it take care of the rest. Returned Completions/Rows are not really relevant. I'm currently missing the ability to let skunk just send any kind of statement, query or command, single or multiple to Postgres without the need to parse the contents or split into single statements.

I currently ended up with adding an extended Session version as in this PR that ships with a "simple"
```scala
def execute_(statement: Statement[Void]): F[Unit]
```
and was wondering whether this or something similar could be considered as part of skunk :thinking: 

Related issue about multi-query statements support: https://github.com/typelevel/skunk/issues/695
This one is maybe also somewhat related https://github.com/typelevel/skunk/issues/959


mpilquist:
This seems like a good idea. Any ideas for a better name than `execute_`?

rolang:
> This seems like a good idea. Any ideas for a better name than `execute_`?

How about `executeDiscard` (or `executeDiscarded`)?
I picked the idea with the `_` from cats where you have methods like `traverse` or `sequence` etc. going in pair with `traverse_` / `sequence_` as versions with discarded output, I thought it may fit as it builds on top of cats.
E.g. similar functionality in ZIO is named like `foreach` / `foreachDiscard`, `foreachPar` / `foreachParDiscard` etc.

rolang:
I guess I could also add a version which accepts arguments if this extension is ok like:
`def executeDiscard[A](statement: Statement[Void])(args: A): F[Unit]` and also try to improve the code coverage.
I was not sure yet what to do with `NoticeResponse` messages, this one is not covered in the test.

rolang:
Update:
- Renamed `execute_` to `executeDiscard`. Another name I just thought of is `executeVoid`, how about that :thinking:  ? 
- Discarding `NoticeResponse` as well instead of failing (not sure it needs to fail on that)
- updated sample queries in tests (something similar I've seen in the "wild" being done in migration scripts)

--- 

> I guess I could also add a version which accepts arguments  

I think I underestimated that one. It would make this pull request somewhat bigger :thinking:  I guess it would need to run through prepare as well and require a new type like `PreparedStatement` along with  `PreparedQuery` / `PreparedCommand`? Maybe it can be done as part of another update.

PR #1026, opened 2024-01-10
Add merge completion case

massimosiani:


mpilquist:
Thanks! Would you mind retargeting this to the series/0.6.x branch?

armanbilge:
Re-triggering CI ...

massimosiani:
I believe that the tests run on pg 11, while MERGE has been introduced in pg 15. Should we run the test conditionally?

mpilquist:
Can we make the test check for merge support and then skip if not enabled on the server?

modules/tests/shared/src/test/scala/CommandTest.scala
```diff
@@ -532,15 +532,19 @@ class CommandTest extends SkunkTest {
...
       _ <- assert("merge", c == Completion.Merge(1))
       c <- s.prepare(selectCity).flatMap(_.option(Garin.id))
       _ <- assert("read", c == None)
+      _ <- s.execute(deleteCity)(Garin.id)
       _ <- s.assertHealthy
-    } yield "ok"
+    } yield "ok")
+      .recoverWith {
+        case SqlState.SyntaxError(ex) if ex.message.startsWith("""Syntax error at or near "MERGE"""") => s.execute(deleteCity)(Garin.id).as("ok")
+      }
```
armanbilge, line 547: Hm, instead of this can we do something like `SELECT version();` to decide whether or not to run this test?

massimosiani: Ah thanks! I was looking for querying direct support of merge, but that's much better of course.

massimosiani:
It doesn't change the discussion above, but PG 11 seems EOL since November 2023 (https://www.postgresql.org/support/versioning/). Should we upgrade the docker to PG 12 (in a separate PR)?

mpilquist:
Yeah that's a good idea.

PR #1048, opened 2024-02-24
Add basic select benchmarks

AlekseiLitkovetc:
Hello @tpolecat,

Thank you so much for such a cool data access library. I see how much effort and time was invested in researching this topic and implementing such a wonderful tool as Skunk. Thank you :pray: 

Recently, I noticed that the Skunk repository doesn't have benchmarks. If you don't mind, can we add some basic selection tests similar to those in the Doobie repository? In case this work is relevant, we can also consider adding more tests (for example, to test DML operations). And I will be glad to help with it.

Thank you in advance!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -257,7 +257,7 @@ jobs:
       - name: Submit Dependencies
         uses: scalacenter/sbt-dependency-submission@v2
         with:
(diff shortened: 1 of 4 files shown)


mpilquist:
This sounds great, thanks!

AlekseiLitkovetc:
Thank you!

PR #1049, opened 2024-02-25
Update copyright

AlekseiLitkovetc:
While adding benchmarks, I noticed that we have outdated copyright. I propose to update it.

If you don't mind, I also added `Contributors`.


tpolecat:
:+1: Yes, should have done this years ago.

PR #1059, opened 2024-03-18
Optimize Parse+Describe with TCP Pipelining

vbergeron:
Second attempt, there I managed to get all the test green locally.

Goal of the PR is to fuse the Parse and the Describe step into one to optimize network exchange.
Despite this being not a very critical workflow, this exchange is simple enough to do a PoC on the concept.

Before:
`Parse -> Flush -> await ParseComplete -> Describe -> Flush -> await ParamDesc -> await RowDesc`

After:
`Parse -> Describe -> Flush -> await ParseComplete -> await ParamDesc -> await RowDesc`

This ensure an optimal usage of the socket and should reduce latencies.


mpilquist:
Thanks!

PR #1061, opened 2024-03-19
TCP Pipelining: Optimize Bind+Execute message flow

vbergeron:
Another pipelining optim query, this time on the Command `bind` + `execute` combo.

Expect a good performance bump when using short-lived command or queries

**Commands**
Before : 
`Bind -> Flush -> await BindComplete -> Execute -> Flush -> await CommandComplete`
After : 
`Bind -> Execute -> Flush -> await BindComplete ->  await CommandComplete`

**Queries with a fixed known size (unique, option, stream)**
Before
* portal creation `Bind -> Flush -> await BindComplete`
* call to execute method `Execute -> Flush -> Unroll`

After:
* portal creation `Bind -> Execute -> Flush -> await BindComplete ->  Unroll and retain the value`
* call to execute method
  * First time : Lookup and use the cached value
  * Following : `Execute -> Flush -> Unroll`

**Queries with an unknown size (cursors)**
No changes

### DIFF
--- a/modules/core/shared/src/main/scala/net/protocol/BindExecute.scala
+++ b/modules/core/shared/src/main/scala/net/protocol/BindExecute.scala
@@ -6,9 +6,10 @@ package skunk.net.protocol
 
 import cats.effect.Resource
 import cats.syntax.all._
-import cats.MonadError
+import cats.effect.Concurrent
+import skunk.~
 import skunk.exception._
-import skunk.net.message.{ Bind => BindMessage, Close => _, _ }
+import skunk.net.message.{ Bind => BindMessage, Execute => ExecuteMessage, Close => _, _ }
 import skunk.net.MessageSocket
 import skunk.net.Protocol.PortalId
 import skunk.util.{ Origin, Namer }
@@ -18,6 +19,7 @@ import skunk.RedactionStrategy
 import skunk.net.Protocol
 import skunk.data.Completion
 import skunk.net.protocol.exchange
+import cats.effect.kernel.Deferred
 
 trait BindExecute[F[_]] {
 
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/net/protocol/BindExecute.scala
```diff
@@ -0,0 +1,186 @@
...
+                        Attribute("portal-id", pn.value)
+                      )
+                _  <- send(ExecuteMessage(pn.value, initialSize))
+                _  <- send(Flush)
+                _  <- postBind
+                rs <- unroll(statement, args, argsOrigin, redactionStrategy)
+              } yield new Protocol.QueryPortal[F, A, B](pn, statement, args, argsOrigin, redactionStrategy) {
+                def execute(maxRows: Int): F[List[B] ~ Boolean] = 
+                  prefetch.tryGet.flatMap {
+                    case None => rs.pure <* prefetch.complete(())
```
mpilquist, line 175: Is it safe to ignore `maxRows` here?

vbergeron: Two use case of this :
* `PreparedQuery.fetch2` : bind with size 2, execute with size 2
* `PreparedQuery.stream`: bind with size `batchSize`, execute with the same size

I resorted to ignore the parameter instead of separating QueryPortal into several cases: one for fixed size query without continuation (fetch2), one for fixed size with continuation (stream), and one for arbitrary sized execute (cursor).

This can be done in a follow up PR to cleanup the internals a bit :)

vbergeron: Moreover, I separated the case `bindSized` from `bind` at the protocol prepared query level. This is clearer to see if we are in the BingExecute case or classic Bind then Execute.

mpilquist (approved):

PR #1065, opened 2024-03-30
Fix #1042 - deadlock in session pool

### LINKED ISSUE #1042: Deadlock while using session pool
I have encountered a deadlock when using the session pool. This happens when I attempt to use multiple sessions from the pool in parallel and the connection to PG fails. I have a minimal example here including logs/fiber dump that shows the waiting fibers in Skunk. If my understanding is correct, reading in parallel should be possible as the pool#use internally uses a semaphore to control permits, but these don’t seem to be being released properly on a connection failure. Could someone please advise if this is an implementation issue, or a bug with the library?

https://github.com/bcarter97/skunk-session-pool-parallel-deadlock

mpilquist:
Fixes #1042. Issue occurred when (a) allocated resource failed to be returned to the pool and (b) resource finalizer also failed with an error.

### DIFF
--- a/modules/core/shared/src/main/scala/util/Pool.scala
+++ b/modules/core/shared/src/main/scala/util/Pool.scala
@@ -152,7 +152,7 @@ object Pool {
           ref.modify {
             case (os, Nil) =>  ((os :+ None, Nil), ().pure[F]) // new empty slot
             case (os, d :: ds) =>  ((os, ds), Concurrent[F].attempt(rsrc(Tracer[F]).allocated).flatMap(d.complete).void) // alloc now!
-          } .guarantee(a._2).flatten
+          }.flatten.guarantee(a._2)
         }
 
       // Hey, that's all we need to create our resource!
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/util/Pool.scala
```diff
@@ -152,7 +152,7 @@ object Pool {
           ref.modify {
             case (os, Nil) =>  ((os :+ None, Nil), ().pure[F]) // new empty slot
             case (os, d :: ds) =>  ((os, ds), Concurrent[F].attempt(rsrc(Tracer[F]).allocated).flatMap(d.complete).void) // alloc now!
-          } .guarantee(a._2).flatten
+          }.flatten.guarantee(a._2)
```
mpilquist, line 155: Upon losing connection to the server, this `dispose` is called. The deferred removed here (`d`) should be completed but it's not b/c the finalizer of `A` completes with the connection error.

Issue #1068, opened 2024-04-09
NumberFormatException when parsing OID values exceeding 32-bit signed integer range


dallinhuff:
## Description:
Currently, `Typer` uses `Integer.parseInt()` for parsing PostgreSQL OIDs. However, PostgreSQL OIDs are 32-bit unsigned integers, and using `Integer.parseInt()` can lead to a `NumberFormatException` for values that exceed the range of a 32-bit signed integer.

## Stack Trace
```
skunk.exception.DecodeException: 
🔥  
🔥  DecodeException
🔥  
🔥    Problem: Decoding error.
🔥     Detail: This query's decoder was unable to decode a row of data.
🔥  
🔥  The statement under consideration was defined
🔥    at «skunk internal»:0
🔥  
🔥    SELECT oid typid, typname, typarray, typrelid
🔥    FROM   pg_type
🔥    WHERE typnamespace IN (
🔥      SELECT oid
🔥      FROM   pg_namespace
🔥      WHERE nspname = ANY(current_schemas(true))
🔥    )
🔥  
🔥  The row in question returned the following values (truncated to 15 chars).
🔥  
🔥    typid     oid   ->  4294967108    ├── java.lang.NumberFormatException (see below)
🔥    typname   name  ->  pg_aggregate                                                 
🔥    typarray  oid   ->  0                                                            
🔥    typrelid  oid   ->  4294967108                                                   
🔥  
🔥  The decoder threw the following exception:
🔥  
🔥    java.lang.NumberFormatException: For input string: "4294967108"
🔥      java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
🔥      java.base/java.lang.Integer.parseInt(Integer.java:662)
🔥      java.base/java.lang.Integer.parseInt(Integer.java:778)
🔥      scala.collection.StringOps$.toInt$extension(StringOps.scala:910)
🔥      skunk.util.Typer$ProtocolOps.$init$$$anonfun$4(Typer.scala:180)
...
```

## Expected behavior:
Typer should parse PostgreSQL OIDs correctly, including values within the 32-bit unsigned integer range.

## Actual behavior:
When encountering a PostgreSQL OID value that exceeds the range of a 32-bit signed integer, `Integer.parseInt()` throws a NumberFormatException, as it cannot handle values outside the signed integer range.

## Proposed solution:
[...]

PR #1086, opened 2024-05-09
Handle Postgres oids greater than Int.MaxValue

benhutchison:
Fix for https://github.com/typelevel/skunk/issues/1068

### DIFF
--- a/modules/core/shared/src/main/scala/util/Typer.scala
+++ b/modules/core/shared/src/main/scala/util/Typer.scala
@@ -177,7 +177,10 @@ object Typer {
   case class TypeInfo(oid: Int, name: String, arrayTypeOid: Option[Int], relOid: Option[Int])
 
   implicit class ProtocolOps[F[_]: Functor](p: Protocol[F]) {
-    val oid: Codec[Int] = Codec.simple(_.toString, _.toInt.asRight, Type.oid)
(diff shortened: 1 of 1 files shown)


modules/core/shared/src/main/scala/util/Typer.scala
```diff
@@ -177,7 +177,10 @@ object Typer {
   case class TypeInfo(oid: Int, name: String, arrayTypeOid: Option[Int], relOid: Option[Int])
 
   implicit class ProtocolOps[F[_]: Functor](p: Protocol[F]) {
-    val oid: Codec[Int] = Codec.simple(_.toString, _.toInt.asRight, Type.oid)
+    //Note Postgres defines oids as *unsigned* Ints https://www.postgresql.org/docs/current/datatype-oid.html
+    //Since Scala currently lacks a built-in unsigned Int type, if the oid exceeds `Int.MaxValue`
+    //it will be converted to/from a negative Int by this Codec (only observed in CockroachDB)
+    val oid: Codec[Int] = Codec.simple(java.lang.Integer.toUnsignedLong(_).toString, _.toLong.toInt.asRight, Type.oid)
```
armanbilge, line 183: I accidentally directly pushed a slight revision of the parsing logic in e66ae4fe87d92d49cc8b5b4d4ebd2365bcc65301.

PR #1114, opened 2024-08-14
WIP: Initial code to test connections before handing them out

non:
The goal is to fix issues where we have dead connections in the pool which are being handed out. In particular, when all the connections in the pool are dead it can easily exhaust a client's retry budget.x


mpilquist:
Had to revert this PR due to a failing PoolTest. See https://github.com/typelevel/skunk/issues/1190

PR #1154, opened 2024-12-20
Support dynamic user/password when building a session pool

### LINKED ISSUE #1109: Support dynamic passwords
**Use Case**
At $work, we're looking to switch from doobie to skunk. One obstacle on the migration comes from the likely requirement to use AWS RDS IAM. It integrates Postgres authentication with AWS IAM by signing short-lived (15 min) tokens that can be used as password when connecting to RDS Postgres. The problem with skunk is that it expects the password to be static, it's just `String` in all relevant signatures.

**Feature request**
Unless I'm overlooking some smart way to deal with a connection pool and a dynamic password in the existing setup, it would be cool if skunk added support. From what I've seen at first glance, taking a `F[String]` (or maybe more general a `F[SomeConnectionParamsIncludingPassword]` that is evaluated when the connection is established, would help to support it. The existing (public) method signatures could just lift the password using `pure`, for RDS IAM, the effect would ask the AWS SDK for a fresh/TTL-aware-cached token.

ant8e:
Closes #1109

### DIFF
--- a/modules/core/shared/src/main/scala/Session.scala
+++ b/modules/core/shared/src/main/scala/Session.scala
@@ -302,7 +302,7 @@ trait Session[F[_]] {
   def closeEvictedPreparedStatements: F[Unit]
 }
 
-
+case class Credentials(user: String, password: Option[String] = none)
 
 /** @group Companions */
 object Session {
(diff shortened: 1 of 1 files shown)


ant8e:
I have been running this in production $WORK for a few weeks, this is working well.

mpilquist (changes requested):
Thanks! Couple minor suggestions and then this should be good for merge.

mpilquist:
Thanks!

PR #1157, opened 2025-01-02
Fixes memory leak in prepared statement cache.  Fixes #1143

### LINKED ISSUE #1143: Postgres memleak caused by prepared statements growing forever when prepareCache is overflowing
Our Postgres db keeps running out of RAM.

We have an infinity stream of events from Kafka, each we groupBy(number = 1000) then go to the db and fetch in batch

We have a few queries which are parametrized with lists like 

sql """ select ...
 where id in (${SomeCodecs.id.list(ids.size)});
"""

We have a session with a pool, prepareCacheSize = 1024
We have about 25 queries 

I think what happens is that we have more prepared statements ( lets say about 25 x 1000 ) than the cache size ,
and when the local cache runs out of space it just disposed it from the local cache but not from the service with deallocate.

When the same query with same number of params to list  comes up, skunk just prepares again since it's no longer in the cache, so the number prepared statement prepared statement on the postgres server just keeps growing until it either dies or we restart the app

There are a few thing I could/will do at our side but I thought it make sense to report this as a bug

matthughes:
Prepared statements were created on the server and cached as part of the "prepare cache."  However, despite entries being removed from the in-memory cache, the evicted entries were not removed from Postgres itself.

This modifies SemispaceCache to keep track of evicted elements.  When a Session is recycled, we close any prepared statements that have been evicted.


modules/core/shared/src/main/scala/Session.scala
```diff
@@ -300,6 +300,10 @@ trait Session[F[_]] {
    */
   def parseCache: Parse.Cache[F]
 
+  /**
+   * Closes any prepared statements evicted as part of this session.  Statements are evicted when the cache limit is reached.
+   */
+  private[skunk] def clearEvicted: F[Unit]
```
matthughes, line 306: I don't love this but I didn't know where else to put it and this solution seemed the most fool-proof implementation with the smallest change to the API.

modules/tests/shared/src/test/scala/SkunkTest.scala
```diff
@@ -15,7 +15,7 @@ import munit.Location
 import org.typelevel.otel4s.trace.Tracer
 import scala.concurrent.duration.Duration
 
-abstract class SkunkTest(debug: Boolean = false, strategy: Typer.Strategy = Typer.Strategy.BuiltinsOnly) extends ffstest.FTest {
+abstract class SkunkTest(debug: Boolean = true, strategy: Typer.Strategy = Typer.Strategy.BuiltinsOnly) extends ffstest.FTest {
```
mpilquist, line 18: Left over true from debugging?

modules/tests/shared/src/test/scala/data/SemispaceCacheTest.scala
```diff
@@ -9,58 +9,4 @@ import org.scalacheck.Gen
 import org.scalacheck.Prop._
 
 class SemispaceCacheTest extends ScalaCheckSuite {
```
mpilquist, line 11: Why remove?

modules/core/shared/src/main/scala/data/SemispaceCache.scala
```diff
@@ -10,20 +10,25 @@ import cats.syntax.all._
...
-    else if (gen0.size < max) SemispaceCache(gen0 + (k -> v), gen1, max) // room in gen0, done!
-    else                      SemispaceCache(Map(k -> v), gen0, max)     // no room in gen0, slide it down
+  def withEvicted(e: List[V]): SemispaceCache[K, V] = 
+    SemispaceCache(gen0, gen1, max, e)
 
-  def lookup(k: K): Option[(SemispaceCache[K, V], V)] =
+  def insert(k: K, v: V): SemispaceCache[K, V] = {
+    if (max == 0)             this.withEvicted(v :: evicted)                                        // special case, can't insert!
+    else if (gen0.size < max) SemispaceCache(gen0 + (k -> v), gen1, max, evicted)                   // room in gen0, done!
+    else                      SemispaceCache(Map(k -> v), gen0, max, gen1.values.toList ::: evicted)// no room in gen0, slide it down
```
mpilquist, line 25: Is there some way we can eagerly close the evicted gen 1 items here? Instead of waiting until session end?

modules/core/shared/src/main/scala/Session.scala
matthughes: Yes, fixing.

modules/tests/shared/src/test/scala/data/SemispaceCacheTest.scala
matthughes: I don't have to but it's completely empty.

mpilquist: Hm, where did those tests go then?

modules/core/shared/src/main/scala/data/SemispaceCache.scala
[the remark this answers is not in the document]
matthughes: Ah I see.  Fixing and writing test for this.

matthughes: I don't know.  I was trying to do that for a while but kept getting stuck on the best place to do it.  I don't like adding a method to Session but it seems the easiest to ensure everything that is evicted is cleaned up.  

For example, if someone used Session.prepare... how would I know when I could safely get rid of that?  It's somewhat easier for the execute methods.

The prepare methods seem to conflict with this cache model.  Before the cache, you had docs saying to prepare the queries ahead of time and then re-use them each session.  But now they are getting cached both manually by the user and in the SemispaceCache.  So you could imagine a bug where user has prepared cache size of 10, then they call

session.prepare { pq =>

  ...
}

Then they create 10 more prepared statements using execute.  This `pq` will get evicted from semispace cache and closed on the Postgres side and when they go to execute it again it will blow up.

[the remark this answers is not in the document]
matthughes: Added fix and test to SemispaceCacheTest

mpilquist: I haven't thought much about it yet but what about something like providing `SemispaceCache` an `onEvicted` callback?

matthughes: Maybe we could say methods that use Session.prepare/prepareR just aren't cached via SemispaceCache and then update the execute/option/stream methods to close any evicted statements immediately?

mpilquist:
@matthughes I pushed a commit which eagerly closes evicted statements. Will add a few review comments with details.

modules/core/shared/src/main/scala/Session.scala
```diff
@@ -365,7 +365,7 @@ object Session {
      * isn't running arbitrary statements then `minimal` might be more efficient.
      */
     def full[F[_]: Monad]: Recycler[F, Session[F]] =
-      closeEvictedPreparedStatements[F] <+> ensureIdle[F] <+> unlistenAll <+> resetAll
```
mpilquist: No longer needed since we now close any evicted statement as soon as possible after cache mutation (either on an insert or get).

modules/core/shared/src/main/scala/data/SemispaceCache.scala
```diff
@@ -12,31 +12,26 @@ import cats.syntax.all._
  */
 sealed abstract case class SemispaceCache[K, V](gen0: Map[K, V], gen1: Map[K, V], max: Int, possiblyEvicted: List[V]) {
 
-  assert(max >= 0)
```
mpilquist, line 15: Requiring max >= 1 allows us to bypass the case where inserting results in immediate eviction. Allowing immediate eviction means the statement ends up getting closed before it can be used.

modules/core/shared/src/main/scala/data/SemispaceCache.scala
```diff
@@ -12,31 +12,26 @@ import cats.syntax.all._
...
     gen1.get(k).map(v => (insert(k, v), v))  // key is in gen1, copy to gen0
-  }
 
   def containsKey(k: K): Boolean =
     gen0.contains(k) || gen1.contains(k)
 
   def values: List[V] =
     (gen0.values.toSet | gen1.values.toSet).toList
+
+  def clearEvicted: (SemispaceCache[K, V], List[V]) =
```
mpilquist, line 33: Combined the accessor and the update in to a single operation.

modules/core/shared/src/main/scala/net/protocol/Close.scala
```diff
@@ -20,27 +20,40 @@ trait Close[F[_]] {
...
-          _ <- send(Flush)
-          _ <- expect { case CloseComplete => }
-        } yield ()
+    new AbstractClose[F] {
+      protected def weakExchange[A](label: String)(f: Span[F] => F[A]) =
+        exchange(label)(f)
+    }
 
+  /** Like [[apply]] but doesn't acquire a mutex, allowing usage from within an existing exchange. */
+  private[skunk] def midExchange[F[_]: FlatMap: MessageSocket: Tracer]: Close[F] =
```
mpilquist, line 29: Total hack to re-use the logic in closing a statement from within the `ParseDescribe` exchange without acquiring the session mutex and hence deadlocking.

modules/core/shared/src/main/scala/net/protocol/ParseDescribe.scala
```diff
@@ -44,14 +44,18 @@ object ParseDescribe {
...
         stmt.encoder.oids(ty) match {
 
           case Right(os) if os.length > Short.MaxValue =>
-            TooManyParametersException(stmt).raiseError[F, (F[StatementId], StatementId => F[Unit])]
+            TooManyParametersException(stmt).raiseError[F, (F[StatementId], StatementId => F[Unit], F[Unit])]
 
           case Right(os) =>
-            OptionT(parseCache.value.get(stmt)).map(id => (id.pure, (_:StatementId) => ().pure)).getOrElse {
+
+            val closeEvictedStatements =
```
mpilquist, line 55: Extracted here b/c this is needed in both the `get` and `insert` cases below (because `get` can cause eviction of `gen1` in case where target entry is in `gen1`).

modules/core/shared/src/main/scala/net/protocol/ParseDescribe.scala
```diff
@@ -70,17 +74,17 @@ object ParseDescribe {
...
-            UnknownTypeException(stmt, err, ty.strategy).raiseError[F, (F[StatementId], StatementId => F[Unit])]
+            UnknownTypeException(stmt, err, ty.strategy).raiseError[F, (F[StatementId], StatementId => F[Unit], F[Unit])]
 
         }
 
       override def command[A](cmd: skunk.Command[A], ty: Typer): F[StatementId] = {
 
         def describeExchange(span: Span[F]): F[(StatementId => F[Unit], F[Unit])] = 
-          OptionT(cache.commandCache.get(cmd)).as(((_: StatementId) => ().pure, ().pure)).getOrElse {
+          OptionT(cache.commandCache.get(cmd)).as(((_: StatementId) => ().pure, cache.commandCache.clearEvicted.void)).getOrElse {
```
mpilquist, line 87: Have to clear evicted cached entries for command cache here (and query cache below) just to ensure we don't leak entries. Unlike statements, we don't need to take any action on the evicted entries -- just ensure they are removed from the eviction space in the underlying `SemispaceCache`.

mpilquist:
Discussed this some today with @matthughes. This PR as-is has a significant flaw -- if a client prepares a statement and holds on to it, then that statement is evicted from the parse cache and subsequently closed on the server, any subsequent use of the evicted statement results in an exception. @matthughes demonstrated this in [8799d74](https://github.com/typelevel/skunk/pull/1157/commits/8799d74fdb1defbd9703997027a14057906cbe14).

One way we could fix is checking if the `StatementId` referenced by a `Protocol.PreparedQuery` is still in the parse cache. If not, then prepare it again and use the resulting statement id instead.

mpilquist:
I restored the logic @matthughes implemented originally, where any evicted prepared statements aren't closed until the recycling of the session they were prepared in. This avoids the risk of a prepared statement being unusable due to eviction during a session.

Only other thing I think we should add is modifying `prepareR` to skip caching altogether and close the prepared statement at resource finalization. Then this should be good to merge.

matthughes:
LGTM

PR #1162, opened 2025-01-14
Fix bug in semispace cache that allowed gen1 to contain values in evi…

matthughes:
…cted.

### DIFF
--- a/modules/core/shared/src/main/scala/data/SemispaceCache.scala
+++ b/modules/core/shared/src/main/scala/data/SemispaceCache.scala
@@ -16,12 +16,12 @@ sealed abstract case class SemispaceCache[K, V](gen0: Map[K, V], gen1: Map[K, V]
   assert(gen0.size <= max)
   assert(gen1.size <= max)
 
-  def insert(k: K, v: V): SemispaceCache[K, V] =
-    if (max == 0)        SemispaceCache(gen0, gen1, max, evicted + v)                      // immediately evict
-    else if (gen0.size < max) SemispaceCache(gen0 + (k -> v), gen1, max, evicted - v)      // room in gen0, done!
-    else                 SemispaceCache(Map(k -> v), gen0, max, evicted ++ gen1.values - v)// no room in gen0, slide it down
+  def insert(k: K, v: V): SemispaceCache[K, V] = 
+    if (max == 0)        SemispaceCache(gen0, gen1, max, evicted + v)                       // immediately evict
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/data/SemispaceCache.scala
```diff
@@ -16,12 +16,12 @@ sealed abstract case class SemispaceCache[K, V](gen0: Map[K, V], gen1: Map[K, V]
   assert(gen0.size <= max)
   assert(gen1.size <= max)
 
-  def insert(k: K, v: V): SemispaceCache[K, V] =
-    if (max == 0)        SemispaceCache(gen0, gen1, max, evicted + v)                      // immediately evict
-    else if (gen0.size < max) SemispaceCache(gen0 + (k -> v), gen1, max, evicted - v)      // room in gen0, done!
-    else                 SemispaceCache(Map(k -> v), gen0, max, evicted ++ gen1.values - v)// no room in gen0, slide it down
+  def insert(k: K, v: V): SemispaceCache[K, V] = 
+    if (max == 0)        SemispaceCache(gen0, gen1, max, evicted + v)                       // immediately evict
+    else if (gen0.size < max) SemispaceCache(gen0 + (k -> v), gen1 - k, max, evicted - v)   // room in gen0, remove from gen1, done!
```
mpilquist, line 21: If `gen0` or `gen1` contains `(k, v0)` such that `v0 != v`, we should add `v0` to evicted list. Likewise below.

modules/core/shared/src/main/scala/data/SemispaceCache.scala
```diff
@@ -40,8 +40,14 @@ sealed abstract case class SemispaceCache[K, V](gen0: Map[K, V], gen1: Map[K, V]
 
 object SemispaceCache {
 
-  private def apply[K, V](gen0: Map[K, V], gen1: Map[K, V], max: Int, evicted: EvictionSet[V]): SemispaceCache[K, V] =
-    new SemispaceCache[K, V](gen0, gen1, max, evicted) {}
+  private def apply[K, V](gen0: Map[K, V], gen1: Map[K, V], max: Int, evicted: EvictionSet[V]): SemispaceCache[K, V] = {
+    val r = new SemispaceCache[K, V](gen0, gen1, max, evicted) {}
+    val gen0Intersection: Set[V] = (gen0.values.toSet intersect evicted.toList.toSet)
```
mpilquist, line 45: Drop these assertions now? I worry a bit about all these set creations/intersections on each constructor?

PR #1163, opened 2025-01-15
Replace SemispaceCache with LRU Cache

mpilquist:
This PR removes `SemispaceCache` and adds a least recently used cache (creatively named `Cache`). This improves cache behavior, especially with access patterns where cached entries were promoted from gen 1 to gen 0 and resultantly lots of cached statements would be evicted.

The LRU implementation is based on maps and access stamps -- not particularly storage efficient but should be good enough for this use case. I thought about implementing this with a splay tree instead but opted for the hash map design.

The configured max cache space is accurately enforced now, whereas `SemispaceCache` would store up to twice the configured max. Hence, the session defaults were all doubled to get the same net maximum.

Eviction tracking is NOT built in to `Cache` like it was in `SemispaceCache`. Instead, putting an element in to a cache returns the element that was evicted as a result, if any. Eviction tracking moved up in to `StatementCache` instead.

cc @matthughes


matthughes (approved):
LGTM.  I built and ran this PR against my integration tests and regular runtime and did not see any issues.  Confirmed that it does not create more pg_prepared_statements than cache limit.

PR #1287, opened 2026-04-19
Update SCRAM client to 3.2

guihigashi:
Bumps `"com.ongres.scram" % "client" % "2.1"` to `"com.ongres.scram" % "scram-client" % "3.2"`. 

This resolves dependency conflicts for users who also have libraries depending on the newer SCRAM client in their classpath (e.g., `"org.postgresql" % "r2dbc-postgresql" % "1.1.1.RELEASE"`). The internal `authenticationSASL` code has been refactored to support the 3.x builder API.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -132,7 +132,7 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
       "com.beachape"  %%% "enumeratum"   % "1.9.0",
     ).filterNot(_ => tlIsScala3.value)
   ).jvmSettings(
-    libraryDependencies += "com.ongres.scram" % "client" % "2.1",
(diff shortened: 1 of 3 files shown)


mpilquist:
Thanks! Instead of upgrading to latest version, it might be better to remove the dependency entirely and make the `Scram.scala` file that's currently used on JS & Native work for JVM too. Mind giving that a try?

guihigashi:
Thanks for reviewing it! I agree that's the better approach, but I don't have the resources to tackle that right now. It would be great as future work, though

mpilquist:
No worries, I made the change in #1288

mpilquist:
Give https://github.com/typelevel/skunk/releases/tag/v1.1.0-RC1 a shot and let me know if it works for you!

PR #1288, opened 2026-04-19
Remove dependency on scram client in favor of native implementation

mpilquist:
Alternative to #1287 that avoids a third party dependency entirely by reusing the JS/Native implementation for JVM.


build.sbt
```diff
@@ -128,14 +131,12 @@ lazy val core = crossProject(JVMPlatform, JSPlatform, NativePlatform)
       "org.typelevel"          %%% "otel4s-semconv-metrics"  % otel4sVersion,
       "org.tpolecat"           %%% "sourcepos"               % "1.2.0",
       "org.typelevel"          %%% "twiddles-core"           % "1.0.0",
+      "com.armanbilge"         %%% "saslprep"                % "0.1.2",
```
armanbilge, line 134: Not urgent, but since it's becoming a core depedency we could move this under `org.typelevel` and 1.0? I can do that.

armanbilge (approved):
awesome that this ended up being useful for the JVM too 😁

build.sbt
mpilquist: Yeah that would be great!

Issue #1276, opened 2026-03-27
Feature Request: Support for Quoted Identifiers


matejcerny:
Currently, Skunk's Identifier models a Postgres Unquoted Identifier:

> *SQL identifiers and key words must begin with a letter or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($)*

However, Skunk currently has no way to represent Postgres Quoted Identifiers:

> *There is a second kind of identifier: the delimited identifier or quoted identifier. It is formed by enclosing an arbitrary sequence of characters in double-quotes ("). (...) Quoted identifiers can contain any character, except the character with code zero.*

Source: [Lexical Structure ](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)

## The Use Case / Blocker

I am currently building a Scala integration for [PGMQ](https://github.com/pgmq/pgmq) (a popular Postgres message queue extension). PGMQ uses LISTEN / NOTIFY channels formatted with dots, for example: `pgmq.q_my_queue.INSERT`. Because this contains a dot (.), it is an invalid unquoted identifier, but a perfectly valid quoted identifier in Postgres.

Since `Identifier` is a `sealed abstract case class`, library authors are completely locked out of constructing these valid quoted identifiers, forced to use raw #$ string interpolation to execute LISTEN commands.

## Proposed Solution

I propose adding a new smart constructor to the Identifier companion object:

```scala
def fromStringQuoted(s: String): Either[String, Identifier] = {
  if (s.length > maxLen)
    Left(s"Identifier too long: ${s.length} (max allowed is $maxLen)")
  else if (s.contains("\u0000"))
    Left("Illegal identifier: cannot contain the null byte (\\u0000).")
  else
    // Wraps the valid string in double quotes so it is sent to Postgres correctly
    Right(new Identifier(s""""$s"""") {})
}
```

Alternatively, if expanding the safe API is not desired, exposing an `unsafeFromString(s: String): Identifier` escape hatch would allow library authors to handle the quoting and validation themselves.

matejcerny:
@mpilquist Could you please take a look 🙏

mpilquist:
Heya apologies on delay. At first glance this seems reasonable to me. Want to give it a shot and open a PR?

mpilquist:
Fixed by #1307

PR #1307, opened 2026-05-13
Support for quoted identifiers

matejcerny:
implements #1276

### DIFF
--- a/modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala
+++ b/modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala
@@ -154,6 +154,21 @@ object StringContextOps {
         return '{???}
     }
 
+  def qidImpl(sc: Expr[StringContext])(using qc: Quotes): Expr[Identifier] =
+    import qc.reflect.report
+    sc match {
+      case '{ StringContext(${Varargs(Exprs(Seq(part)))}: _*) } =>
+        Identifier.fromStringQuoted(part) match {
+          case Right(Identifier(s)) => '{ Identifier.fromStringQuoted(${Expr(s)}).fold(sys.error, identity) }
+          case Left(s) =>
+            report.error(s)
+            return '{???}
+        }
+      case _ =>
+        report.error(s"Identifiers cannot have interpolated arguments")
+        return '{???}
+    }
+
 }
 
 trait ToStringContextOps {
@@ -163,6 +178,9 @@ trait ToStringContextOps {
 
   extension (inline sc: StringContext) inline def id(): Identifier =
     ${ StringContextOps.idImpl('sc) }
+
+  extension (inline sc: StringContext) inline def qid(): Identifier =
+    ${ StringContextOps.qidImpl('sc) }
 
   implicit def toStringOps(sc: StringContext): StringContextOps =
     new StringContextOps(sc)
--- a/modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala
+++ b/modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala
@@ -143,8 +143,8 @@ object StringContextOps {
     import qc.reflect.report
     sc match {
       case '{ StringContext(${Varargs(Exprs(Seq(part)))}: _*) } =>
-        Identifier.fromString(part) match {
-          case Right(Identifier(s)) => '{ Identifier.fromString(${Expr(s)}).fold(sys.error, identity) }
+        Identifier.fromStringLegacy(part) match {
+          case Right(Identifier(s)) => '{ Identifier.fromValue(${Expr(s)}).fold(sys.error, identity) }
           case Left(s) =>
             report.error(s)
             return '{???}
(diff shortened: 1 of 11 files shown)


modules/core/shared/src/main/scala/data/Identifier.scala
```diff
@@ -9,7 +9,15 @@ import cats.Eq
 import scala.util.matching.Regex
 
 sealed abstract case class Identifier(value: String) {
-  override def toString: String = value // ok?
+  def quoted: Boolean = false
+  def asSql: String = if (quoted) s"\"${value.replace("\"", "\"\"")}\"" else value
+  override def toString: String = asSql
+
+  override def equals(other: Any): Boolean = other match {
```
matejcerny: Not sure if this can break anything?

modules/core/shared/src/main/scala/net/message/package.scala
```diff
@@ -58,7 +58,7 @@ package object message { module =>
 
   val identifier: SCodec[Identifier] =
     utf8z.exmap(
-      s  => Attempt.fromEither(Identifier.fromString(s).leftMap(Err(_))),
+      s  => Attempt.fromEither(Identifier.fromString(s).orElse(Identifier.fromStringQuoted(s)).leftMap(Err(_))),
```
matejcerny: Any idea if there is a better solution?

matejcerny:
It's a much bigger change than I originally thought 🤷

modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala
```diff
@@ -154,6 +154,21 @@ object StringContextOps {
...
     }
 
+  def qidImpl(sc: Expr[StringContext])(using qc: Quotes): Expr[Identifier] =
+    import qc.reflect.report
+    sc match {
+      case '{ StringContext(${Varargs(Exprs(Seq(part)))}: _*) } =>
+        Identifier.fromStringQuoted(part) match {
+          case Right(Identifier(s)) => '{ Identifier.fromStringQuoted(${Expr(s)}).fold(sys.error, identity) }
+          case Left(s) =>
+            report.error(s)
```
kubukoz, line 164: there's a variant that aborts and returns `Nothing`, so you don't have to do `'{???}` - `errorAndAbort`

modules/core/shared/src/main/scala/data/Identifier.scala
kubukoz: why not add `quoted` as a constructor parameter instead and rely on the default one?

matejcerny: that would break MiMa

kubukoz: it's a sealed class, it should be fine...

(but maybe it's not)

matejcerny: It's not 😞

modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala
matejcerny: I mirrored the `identifier_impl` implementation

modules/core/shared/src/main/scala/Channel.scala
```diff
@@ -120,29 +120,29 @@ object Channel {
...
         _      <- Resource.make(listen)(_ => unlisten)
         stream <- proto.notifications(maxQueued)
-      } yield stream.filter(_.channel === name)
+      } yield stream.filter(_.channel.value === name.value)
 
 
     def notify(message: String): F[Unit] =
       // TODO: escape the message
-      proto.execute(Command(s"NOTIFY ${name.value}, '$message'", Origin.unknown, Void.codec)).void
+      proto.execute(Command(s"NOTIFY ${name.asSql}, '$message'", Origin.unknown, Void.codec)).void
```
nigredo-tori: IMO we need to support `Identifier`s in interpolation wherever `Fragment[Void]` is allowed. But that's probably a separate issue.

modules/core/shared/src/main/scala/data/Identifier.scala
```diff
@@ -9,7 +9,15 @@ import cats.Eq
 import scala.util.matching.Regex
 
 sealed abstract case class Identifier(value: String) {
-  override def toString: String = value // ok?
+  def quoted: Boolean = false
+  def asSql: String = if (quoted) s"\"${value.replace("\"", "\"\"")}\"" else value
```
nigredo-tori: Do we actually need to treat quoted identifiers as a separate case? As I understand it (could anyone more familiar with PostgreSQL please confirm this?), the distinction exists only on syntactical level:
1. Unquoted identifiers are automatically converted to lowercase.
2. Unquoted identifiers can not contain keywords.
3. Quoted identifiers are not converted, and can contain almost anything.

Which means that `foo_bar` is for all intents and purposes the same as `"foo_bar"`.

I propose we do the following (starting from `main`):
1. Change the existing `id"..."` syntax to automatically lowercase the literal. Maybe mark this syntax as deprecated, since the same behavior can be achieved if the calling code provides the lowercase literal.
2. Add a separate syntax that does not change the literal. ( Not sure what to callit. `qid"..."` doesn't really work, since quoting becomes an implementation detail).
3. When interpolating an `Identifier` into SQL, quote it iff it requires quoting (if the value contains uppercase letters, or otherwise breaks the constraints on unquoted identifiers). We can maybe cache the possibly quoted form on `Identifier` creation.
4. Instead of 3, we can just always quote `Identifier`s. It's noisy WRT generated SQL, but a bit easier to implement.

mpilquist: Do we want to pursue this alternative approach instead of merging this as-is?

matejcerny: Yes, I almost have it ready. (if we agree that deprecating `id"..."` is fine - to me it makes sense)

matejcerny: @nigredo-tori addressed in 16387907f8a3fac8875f41c549b44ff81e04c23d, thanks for the suggestions!

mpilquist:
Thanks!

PR #1323, opened 2026-07-03
Fail notification streams when the connection is lost

rpiaggio:
`Session#channel(...).listen(...)` (and `Protocol#notifications`) hanged silently forever if the session's connection dies. 

The cause was in `BufferedMessageSocket`: the background read loop's failure handler only offered a `NetworkError` poison pill to the synchronous message queue, which raised on the next `send`/`receive` exchange. A LISTEN-only session never performs another exchange, and the notifications `Topic` was neither failed nor closed, so its subscribers waited forever.

This was found and fixed by Claude while looking for the root cause of subscribers not reporting changes on a running server and the reinitialization logic not being triggered.

### DIFF
--- a/modules/core/shared/src/main/scala/net/BufferedMessageSocket.scala
+++ b/modules/core/shared/src/main/scala/net/BufferedMessageSocket.scala
@@ -94,10 +94,11 @@ object BufferedMessageSocket {
    */
   private def next[F[_]: MonadThrow](
     ms:    MessageSocket[F],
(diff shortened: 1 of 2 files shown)


tpolecat:
@mpilquist Hey can we merge this and cut a release? It's been so long I don't know what your release process is.

mpilquist:
Yep want to get this & the Otel changes in - will get a release out today or tomorrow.

PR #1324, opened 2026-07-04
Add _uuid (UUID array) codec

### LINKED ISSUE #1310: Add _uuid array codec
It would be useful if Skunk exposed a built-in `_uuid` codec alongside uuid, similar to the existing array codecs like `_int8` and `_text`.
I use this for batching queries with `WHERE id = ANY($1::uuid[])` and this currently needs either dynamic `uuid.list(...)` SQL or a small custom `Codec.array[UUID](...)`, while scalar UUIDs are already supported via `uuid`.
This would make UUID batch queries consistent with the existing numeric/text array codec APIs.

nigredo-tori:
Fixes #1310

### DIFF
--- a/modules/core/shared/src/main/scala/codec/UuidCodec.scala
+++ b/modules/core/shared/src/main/scala/codec/UuidCodec.scala
@@ -12,14 +12,16 @@ import skunk.data.Arr
 
 trait UuidCodec {
 
-  private def parse(s: String): Either[String, UUID] =
-      Either.catchOnly[IllegalArgumentException](UUID.fromString(s)).leftMap(_.getMessage)
-
   val uuid: Codec[UUID] =
-    Codec.simple[UUID](_.toString, parse, Type.uuid)
+    Codec.simple[UUID](_.toString, codec.uuid.parse, Type.uuid)
 
-  val _uuid: Codec[Arr[UUID]] =
-    Codec.array[UUID](_.toString, parse, Type._uuid)
+  def _uuid: Codec[Arr[UUID]] = codec.uuid._uuidImpl
 }
 
-object uuid extends UuidCodec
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/codec/UuidCodec.scala
```diff
@@ -8,16 +8,18 @@ package codec
 import cats.syntax.all._
 import java.util.UUID
 import skunk.data.Type
+import skunk.data.Arr
 
 trait UuidCodec {
```
nigredo-tori, line 13: The name is now slightly wrong because we have multiple codecs, but it doesn't seem worth fixing while keeping binary compatibility.

modules/core/shared/src/main/scala/codec/UuidCodec.scala
```diff
@@ -8,16 +8,20 @@ package codec
...
 
   val uuid: Codec[UUID] =
-    Codec.simple[UUID](
-      u => u.toString,
-      s => Either.catchOnly[IllegalArgumentException](UUID.fromString(s)).leftMap(_.getMessage),
-      Type.uuid
-    )
+    Codec.simple[UUID](_.toString, codec.uuid.parse, Type.uuid)
 
+  def _uuid: Codec[Arr[UUID]] = codec.uuid._uuidImpl
```
nigredo-tori, line 18: Sadly, we can't add values directly to trait lest MiMa complains. Does anyone have a better way to do this?

nigredo-tori:
I'm not quite sure what to make of the Native test failing, since I have literally zero Native experience.

mpilquist:
Thanks! Native build issue was unrelated to this PR and fixed on main since.

PR #1329, opened 2026-07-09
Add new completion cases for `SERVER` & `USER MAPPING`

TarasZhere:
This change adds support for foreign data wrapper _(FDW)_ related commands. It introduces new Completion cases for `CREATE`/`ALTER`/`DROP SERVER` and `CREATE`/`ALTER`/`DROP USER MAPPING`. The `CommandComplete` message handling is updated to recognize these new completion tags from PostgreSQL. A few new test case have been added to verify the execution of these FDW commands.


mpilquist:
Thank you!

PR #1340, opened 2026-07-16
Support `Identifier` in `sql` interpolation

### LINKED ISSUE #1337: Support `Identifier` in `sql` interpolation
`Identifier` represents a name of a PosgreSQL object (table, column, index...). As far as I can tell, its main purpose is to be interpolated verbatim into SQL. See, for example, [how it is used in `Channel`](https://github.com/typelevel/skunk/blob/59f4a0112a5bb9a06b721012d5832ebacb589898/modules/core/shared/src/main/scala/Channel.scala#L123). Thus I would expect that interpolation is as streamlined as possible:

```scala
val table = ident"foo"
val fragment = sql"SELECT bar FROM $table"
```

Instead, we're expected to use the verbatim interpolation syntax (noisy) and conversion to `String` (noisy and smelly):

```scala
val table = ident"foo"
val fragment = sql"SELECT bar FROM #${table.sql}"
```

I propose we handle `Identifier`s as just another case in `sql` interpolation (e.g. [here](https://github.com/typelevel/skunk/blob/59f4a0112a5bb9a06b721012d5832ebacb589898/modules/core/shared/src/main/scala-3/syntax/StringContextOps.scala#L94-L108)), such that the first example above compiles and works as expected.

### Verbatim syntax

We could still require verbatim syntax for clarity:

```scala
val table = ident"foo"
val fragment = sql"SELECT bar FROM #$table"
```

Personally, I don't think this is warranted, since we already can interpolate `Encoder`s potentially containing arbitrary SQL. I think we should keep the verbatim syntax for **dangerous** cases that could potentially result in SQL injection. Identifiers (as long as properly implemented) do not pose such a threat.

### Other interpolators

It would make sense to allow interpolating both `Identifier`s and `Fragment[Void]`s into `const"..."` and `void"..."`.  But, since interpolating `Fragment[Void]` there is not currently supported, that should probably be left to a future interpolator overhaul.

nigredo-tori:
Closes #1337

### DIFF
--- a/modules/tests/shared/src/test/scala/FragmentTest.scala
+++ b/modules/tests/shared/src/test/scala/FragmentTest.scala
@@ -31,8 +31,18 @@ class FragmentTest extends SkunkTest {
     }
   }
 
-  sessionTest("~") { s =>
+  sessionTest("*:") { s =>
     val f = sql"select $int4" *: sql", $varchar"
+    s.prepare(f.query(int4 *: varchar)).flatMap { ps =>
+      for {
+        n <- ps.unique((123, "456"))
+        _ <- assertEqual("(123, \"456\")", n, 123 *: "456" *: EmptyTuple)
+      } yield "ok"
+    }
+  }
+
+  sessionTest("~") { s =>
+    val f = sql"select $int4" ~ sql", $varchar"
     s.prepare(f.query(int4 *: varchar)).flatMap { ps =>
       for {
         n <- ps.unique((123, "456"))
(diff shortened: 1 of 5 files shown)


modules/tests/shared/src/test/scala/StringContextOpsTest.scala
```diff
@@ -0,0 +1,25 @@
...
+// For more information see LICENSE or https://opensource.org/licenses/MIT
+
+package tests
+package data
+
+import skunk.implicits._
+import skunk.Fragment
+import skunk.Void
+
+class StringContextOpsTest extends ffstest.FTest {
```
nigredo-tori, line 12: I haven't found a separate test file for the syntax, so I've added a new one.

modules/tests/shared/src/test/scala/FragmentTest.scala
```diff
@@ -31,7 +31,7 @@ class FragmentTest extends SkunkTest {
     }
   }
 
-  sessionTest("~") { s =>
+  sessionTest("*:") { s =>
```
nigredo-tori, line 34: This used to check the twiddles' `*:` syntax instead of skunk's `~`. I've changed the name and added another test for `~`. This is unrelated to the main focus of this PR, but, IMO, this doesn't warrant a separate one.

PR #1352, opened 2026-08-27
Use `TracerProvider` and `MeterProvider` instead of `Tracer` and `Meter`

iRevive:
## Motivation

OpenTelemetry recommends libraries to **own** their instrumentation scopes. 
A few benefits:
1) Full control of the namespace - the traces & metrics will belong to the `org.typelevel.skunk` - which simplifies the filtering
2) Every release will be reflected in the instrumentation scope version - much easier to spot regressions and filter traces/metrics if needed

____

https://opentelemetry.io/docs/concepts/instrumentation-scope/

> For libraries and frameworks: use the library’s fully qualified name and version as the scope. If you are writing an instrumentation library for a library that has no built-in OpenTelemetry support, use the name and version of the instrumentation library itself.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,7 +21,13 @@ ThisBuild / githubWorkflowOSes := Seq("ubuntu-latest")
 ThisBuild / githubWorkflowJavaVersions := Seq(JavaSpec.temurin("17"))
 ThisBuild / tlJdkRelease := Some(8)
 ThisBuild / resolvers += Resolver.sonatypeCentralSnapshots
-ThisBuild / githubWorkflowBuildPreamble ++= nativeBrewInstallWorkflowSteps.value
+ThisBuild / githubWorkflowBuildPreamble ++= Seq(
+  WorkflowStep.Run(
+    commands = List("/home/linuxbrew/.linuxbrew/bin/brew update"),
+    name = Some("Update brew"),
+    cond = Some("startsWith(matrix.os, 'ubuntu')")
+  )
+) ++ nativeBrewInstallWorkflowSteps.value
 ThisBuild / nativeBrewInstallCond := Some("matrix.project == 'skunkNative'")
 
 lazy val setupCertAndDocker = Seq(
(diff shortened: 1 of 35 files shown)


modules/core/shared/src/main/scala/Session.scala
```diff
@@ -885,45 +824,6 @@ object Session {
       .withParseCacheSize(parseCache)
       .single
 
-  /**
-   * Resource yielding logically unpooled sessions given a Tracer. This can be convenient for demonstrations and
-   * programs that only need a single session. In reality each session is managed by its own
-   * single-session pool.
-   * @see pooledF
-   */
-  @deprecated("Use Session.Builder[F].singleExplicitTracer instead", "1.0.0-M11")
```
iRevive, line 894: If we want to keep it, we need to change the signature to:
```scala
def singleF[F[_]: Temporal: Network: Console](
  ...
): Telemetry[F] => Resource[F, Session[F]]
```

build.sbt
```diff
@@ -21,7 +21,13 @@ ThisBuild / githubWorkflowOSes := Seq("ubuntu-latest")
...
 ThisBuild / tlJdkRelease := Some(8)
 ThisBuild / resolvers += Resolver.sonatypeCentralSnapshots
-ThisBuild / githubWorkflowBuildPreamble ++= nativeBrewInstallWorkflowSteps.value
+ThisBuild / githubWorkflowBuildPreamble ++= Seq(
+  WorkflowStep.Run(
+    commands = List("/home/linuxbrew/.linuxbrew/bin/brew update"),
+    name = Some("Update brew"),
+    cond = Some("startsWith(matrix.os, 'ubuntu')")
+  )
+) ++ nativeBrewInstallWorkflowSteps.value
```
iRevive, line 30: Hm, I'm starting to think the `update` step is not necessary, since other branches somehow work 🤔

project/plugins.sbt
```diff
@@ -11,3 +11,4 @@ addSbtPlugin("org.scoverage"      % "sbt-scoverage"
 addSbtPlugin("org.scala-js"       % "sbt-scalajs"                                 % "1.22.0")
 addSbtPlugin("org.scala-native"   % "sbt-scala-native"                            % "0.5.12")
 addSbtPlugin("com.armanbilge"     % "sbt-scala-native-config-brew-github-actions" % "0.4.0")
+addSbtPlugin("com.eed3si9n"       % "sbt-buildinfo"                               % "0.13.1")
```
mpilquist, line 14: Is this needed?

mpilquist: Nevermind I get it :)

PR #1354, opened 2026-08-28
Fix to not hand out errored/closed sessions from pool

### LINKED ISSUE #524: Broken Pipe Exception from Pool

### LINKED ISSUE #1190: Un-revert #1114
I had to revert #1114 on main due to a test failure in PoolTest:

```
==> X tests.PoolTest.progress and safety with many fibers and reset failures  0.044s skunk.util.Pool$ResourceLeak:
🔥
🔥  ResourceLeak
🔥
🔥    Problem: A resource leak was detected during pool finalization.
🔥     Detail: Expected 10 active slot(s) and no deferrals, found 107 slots and 0
🔥             deferral(s).
🔥       Hint: The most common causes of resource leaks are (a) using a pool on a
🔥             fiber that was neither joined or canceled prior to pool finalization,
🔥             and (b) using `Resource.allocated` and failing to finalize allocated
🔥             resources prior to pool finalization.
🔥

skunk.util.Pool$ResourceLeak: A resource leak was detected during pool finalization.
    at skunk.util.Pool$.$anonfun$ofF$21(Pool.scala:186)
    at cats.Applicative.whenA(Applicative.scala:264)
    at cats.Applicative.whenA$(Applicative.scala:263)
    at void @ org.typelevel.keypool.KeyPool$.reap(KeyPool.scala:187)
    at *>$extension @ org.typelevel.keypool.internal.RequestSemaphore$.apply(RequestSemaphore.scala:79)
    at *>$extension @ org.typelevel.keypool.internal.RequestSemaphore$.apply(RequestSemaphore.scala:79)
    at get @ org.typelevel.otel4s.sdk.trace.processor.BatchSpanProcessor.org$typelevel$otel4s$sdk$trace$processor$BatchSpanProcessor$$exportAll(BatchSpanProcessor.scala:160)
    at flatMap @ skunk.util.Pool$.free$1(Pool.scala:178)
    at foreverM$extension @ org.typelevel.otel4s.sdk.trace.processor.BatchSpanProcessor.org$typelevel$otel4s$sdk$trace$processor$BatchSpanProcessor$$worker(BatchSpanProcessor.scala:100)
    at mapN @ org.http4s.ember.client.EmberConnection.isValid(EmberConnection.scala:50)
    at unsafeToFuture @ munit.CatsEffectSuite$$anonfun$1.applyOrElse(CatsEffectSuite.scala:82)
```

Revert was done in 932002727c0aacb117217c489614211d2728235. Let's un-revert that and determine what needs fixing.

vlovgr:
This pull request should fix https://github.com/typelevel/skunk/issues/524. Closes #1190.

The `"disconnect while idle in pool"` test shows the issue and fails with the following exception in f0484dff591b78a3283ec428cd4d80737e7db740.

```
🔥  EofException
🔥
🔥    Problem: EOF was reached on the network socket.
🔥     Detail: Attempt to read 5 byte(s) failed after 0 bytes(s) were read, because
🔥             the connection had closed.
🔥       Hint: Discard this session and retry with a new one.
🔥
```

The same test passes in 05ccbb646f34594b28a0ac59a81c806a58ed5ebc.

The issue is that a session can error/close (properly) while being idle in the pool. There's currently no check that this happens, and hence we get the exception above. This is especially common when using a connection pooler like RDS Proxy or PgBouncer, which closes connections after a certain idle timeout.

The solution in this pull request is to detect when an error/close happens in `BufferedMessageSocket`, exposed via the new `isHealthy`. This works well because we continue reading on the socket even while the connection is idle. The `isHealthy` function is propagated through `Protocol#isHealthy` and `Session#isHealthy`. We add a new `Recyclers.ensureHealthy` using this, and a new `Pool.ofF` variant which also accepts a checkout `Recycler` (used when sessions are retrieved from the pool). Finally, we update `Session` to use the `Recyclers.ensureHealthy` as a checkout recycler, which fixes the issue.


modules/core/shared/src/main/scala/net/Protocol.scala
```diff
@@ -280,8 +286,14 @@ object Protocol {
           protocol.Startup[F](opDuration).apply(user, database, password, parameters)
 
         override def cleanup: F[Unit] =
-          parseCache.value.values.flatMap(_.traverse_(protocol.Close[F](opDuration).apply))
-      
+          isHealthy.ifM(
+            parseCache.value.values.flatMap(_.traverse_(protocol.Close[F](opDuration).apply)),
```
vlovgr: There should be no point in sending close here, since the connection has already errored/closed. The test in `QueryTest` has been updated to reflect this.

modules/core/shared/src/main/scala/Session.scala
```diff
@@ -412,21 +420,28 @@ object Session {
      * isn't running arbitrary statements then `minimal` might be more efficient.
      */
     def full[F[_]: Monad]: Recycler[F, Session[F]] =
-      closeEvictedPreparedStatements[F] <+> ensureIdle[F] <+> unlistenAll <+> resetAll
+      ensureHealthy[F] <+> closeEvictedPreparedStatements[F] <+> ensureIdle[F] <+> unlistenAll <+> resetAll
```
vlovgr, line 423: Adding `ensureHealthy` here and to `minimal` is a cheap check to avoid additional work.

PR #1355, opened 2026-08-31
otel4s: introduce Telemetry

iRevive:
This PR is based on https://github.com/typelevel/skunk/pull/1352.

## Key changes

- First of all, we have `TelemetryConfig` which allows customization. Secondly, `Telemetry` is a centralized telemetry hub. 
- Metrics and spans are mostly aligned with https://opentelemetry.io/docs/specs/semconv/db/postgresql/
- Users can attach custom attributes to a query

___

I think we can split this PR into smaller pieces if needed.

___

Some boilerplate changes were implemented by codex.


modules/core/shared/src/main/scala/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,290 @@
...
+      queryText
+        .filter(_.nonEmpty)
+        .foreach(q => span += DbAttributes.DbQueryText(q))
+    } else if (queryTextPolicy == QueryCaptureConfig.QueryTextPolicy.UnsafeAlways) {
+      span += DbAttributes.DbQueryText(statement.sql)
+    }
+
+    config.captureQuery.queryParametersPolicy match {
+      case QueryCaptureConfig.QueryParametersPolicy.All =>
+        redactionStrategy.redactArguments(arguments).zipWithIndex.foreach {
```
iRevive, line 267: First, we redact arguments when we build a query; then we redact them again here. 

The same has been happening before my changes. 

Any reason to try to redact arguments once?

modules/core/shared/src/main/scala/telemetry/Telemetry.scala
```diff
@@ -0,0 +1,290 @@
...
+import org.typelevel.otel4s.semconv.attributes.{DbAttributes, ErrorAttributes, ServerAttributes}
+import org.typelevel.otel4s.semconv.metrics.DbMetrics
+import org.typelevel.otel4s.trace.{SpanFinalizer, SpanKind, StatusCode, Tracer, TracerProvider}
+import skunk.data.Encoded
+import skunk.exception.PostgresErrorException
+import skunk.{BuildInfo, RedactionStrategy, Statement}
+
+import scala.util.control.NonFatal
+
+sealed trait Telemetry[F[_]] {
```
iRevive, line 26: Hm, I think in the current setup we can make it `private[skunk]`. `Session` no longer exposes/requires the `Telemetry` type in the public API.

iRevive: Ah, `Pool` exposes `Telemetry` in the public API. So it must remain public.

modules/core/shared/src/main/scala/telemetry/Telemetry.scala
mpilquist: No, I think when redaction was initially added, there wasn't a great place to do it once

